PR #3, opened 2017-06-15
K8s ready state check

prabalsharma:


cf-deploy-teardown/tasks/k8s-ready-state-check.sh
```diff
@@ -0,0 +1,59 @@
+#Script to determine is the K8s host is "ready" for cf deployment
```
alexjh: Should have a `#!` to set which shell is being used here.

cf-deploy-teardown/tasks/k8s-ready-state-check.sh
```diff
@@ -0,0 +1,59 @@
...
+
+set -ex
+
+# cgroup memory & swap accounting in /proc/cmdline
+
+cat /proc/cmdline | grep -w "cgroup_enable=memory"
```
alexjh: `cat` isn't needed here or in the next check: `grep -w "cgroup_enable=memory" /proc/cmdline` instead.

cf-deploy-teardown/tasks/k8s-ready-state-check.sh
alexjh: Looks like it uses some `bash`isms below like `[[`

alexjh:
If this script is for checking that the k8s cluster is ready, why is it in the `cf-deploy-teardown` dir?

cf-deploy-teardown/tasks/k8s-ready-state-check.sh
```diff
@@ -0,0 +1,59 @@
...
+echo "Verified: cgroug_enable memory"
```
alexjh: s/cgroug/cgroup

cf-deploy-teardown/tasks/k8s-ready-state-check.sh
prabalsharma: oops forgot in copy paste

cf-deploy-teardown/k8s-ready-state-check.sh
```diff
@@ -0,0 +1,62 @@
+# Script to determine is the K8s host is "ready" for cf deployment
```
alexjh: This file should be `chmod +x` so it can run directly from a checkout.

prabalsharma:
>If this script is for checking that the k8s cluster is ready, why is it in the cf-deploy-teardown dir?

Because, this script need to run before cf deploy script and other pieces like script to create K8s host or cf pipeline are not yet ready. I placed it here as this is the only place where it could fit for now. Might need to be moved later when k8s host creation script gets ready or we will run it from this place itself on k8s host before deploying CF

cf-deploy-teardown/tasks/k8s-ready-state-check.sh
```diff
@@ -0,0 +1,61 @@
...
+fi
+
+kubelet=$(systemctl status kubelet -l | grep "/usr/bin/hyperkube kubelet" )
+if [[ $kubelet == *"--allow-privileged"* ]]; then
+  echo "Verified: privileged enabled in 'kubelet'"
+fi
+
+# dns check for the current hostname resolution
+
+IP=$(nslookup cf-dev.io | grep answer: -A 2 | grep Address: | sed 's/Address: *//g')
```
alexjh: Could this be done now, turning `cf-dev.io` into a variable that's set by the CI system? Ie `${${CF_DEPLOYMENT_HOSTNAME}:-cf-dev.io}`  (I didn't test that, just an idea)

cf-deploy-teardown/tasks/k8s-ready-state-check.sh
```diff
@@ -0,0 +1,61 @@
...
+
+kube_dns=$(k get pods : | grep "kube-dns-")
+if [[ $kube_dns == *"4/4 Running"* ]]; then
+  echo "Verified: 'kube-dns shows 4/4 ready'"
+fi
+
+# ntp is installed and running
+
+systemctl status ntpd| grep -w "Active: active (running)"
+echo "Verified: ntp is running"
```
alexjh: My vagrant box doesn't show ntp running... should it?

cf-deploy-teardown/tasks/k8s-ready-state-check.sh
prabalsharma: mmm I left it like this because its not clear how Hart is going to implement this. Either it will just be hostname.ci.van which can easily be picked by command "hostname" (no need for it to be a input parameter/ variable) or it will remain cf-dev.io. May be its better to do a final implementation once we have something related to k8s host creation for which this script actually is. But if you think${${CF_DEPLOYMENT_HOSTNAME}:-cf-dev.io} is a better way, i am okay to do that as well :)

cf-deploy-teardown/tasks/k8s-ready-state-check.sh
prabalsharma: nope. its for K8 host, Aaron said it does no matter on Vagrant box. So i just used `sudo systemctl enable ntpd; sudo systemctl start ntpd` to test this

alexjh: It looks like systemd has a command for checking this already: `is-active`.

```bash
vagrant@vagrant-kube:~/scf> systemctl is-active ntpd
active
vagrant@vagrant-kube:~/scf> echo $?
0
vagrant@vagrant-kube:~/scf> sudo systemctl stop ntpd
vagrant@vagrant-kube:~/scf> systemctl is-active ntpd
unknown
vagrant@vagrant-kube:~/scf> echo $?
3
```

cf-deploy-teardown/tasks/k8s-ready-state-check.sh
alexjh: Oh, I see what you mean. Ok, let's leave it.

PR #4, opened 2017-06-20
Matched name of pipeline to SCF/UFR target/script names.

andreas-kupries:
https://trello.com/c/eX0agPsP/189-clean-up-the-names-of-makefile-targets-and-scripts
Related PRs
- https://github.com/SUSE/scf/pull/1006
- https://github.com/SUSE/uaa-fissile-release/pull/25

### DIFF
--- a/cf-kube-dist/cf-kube-dist.yml
+++ b/cf-kube-dist/cf-kube-dist.yml
@@ -5,8 +5,7 @@ resources:
   type: git
   source:
     branch: develop
-    private_key: {{github-private-key}}
-    uri: git@github.com:hpcloud/hcf.git
+    uri: https://github.com/SUSE/scf.git
 
 - name: ci
   type: git
@@ -14,7 +13,7 @@ resources:
     uri: https://github.com/SUSE/cf-ci.git
     branch: master
     paths:
-    - cf-kube-config/*
+    - cf-kube-dist/*
 
 - name: s3.fissile-binary
   type: s3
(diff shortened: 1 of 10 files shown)


cf-kube-dist/cf-kube-dist.yml
```diff
@@ -6,15 +6,15 @@ resources:
   source:
     branch: develop
     private_key: {{github-private-key}}
```
mook-as: That means we no longer need the private key, right? (Since we're not using ssh anymore)

andreas-kupries:
Looks like I missed a `kube-config`
See also https://github.com/SUSE/uaa-fissile-release/pull/24
in this set.

cf-kube-dist/cf-kube-dist.yml
andreas-kupries: I guess so ?

andreas-kupries:
Addressed in commit 7358fe4d50a3d5206cf2f2e416cc51bc6d52df64

cf-kube-dist/cf-kube-dist.yml
andreas-kupries: Addressed in commit c4b6035602bb93e45fcb73f18f38db346a118220

andreas-kupries:
Should be fixed with commit 0c9e2697f6adc1ebc608105962a30694e1176138

PR #7, opened 2017-06-26
fix k8s check script

prabalsharma:
fix dns check, fix kube-dns check, remove grep for type or kind on persistent storage class check, script can now wither take domain name from env variables or can use the default cf-dev.io

### DIFF
--- a/cf-deploy-teardown/tasks/k8s-ready-state-check.sh
+++ b/cf-deploy-teardown/tasks/k8s-ready-state-check.sh
@@ -3,6 +3,7 @@
 #Script to determine is the K8s host is "ready" for cf deployment
 FAILED=0
 
+SCF_DOMAIN=${SCF_DOMAIN:-cf-dev.io}
 function green() {
   awk '{ print "\033[32mVerified: " $0 "\033[0m" }';
 }
@@ -33,8 +34,8 @@ status "docker info should show overlay2"
 
 # kube-dns shows 4/4 ready
 
-kube_dns=$(kubectl get pods --all-namespaces | grep -q "kube-dns-")
-[[ $kube_dns == *"4/4 Running"* ]]
+kube_dns=$(kubectl get pods --all-namespaces | grep "kube-dns-")
+[[ $kube_dns =~ 4/4\ *Running ]]
 status "kube-dns should shows 4/4 ready"
 
 # ntp is installed and running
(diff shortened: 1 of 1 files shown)


cf-deploy-teardown/tasks/k8s-ready-state-check.sh
```diff
@@ -59,8 +60,7 @@ status "Privileged must be enabled in 'kubelet'"
 
 # dns check for the current hostname resolution
 
-IP=$(nslookup cf-dev.io | grep answer: -A 2 | grep Address: | sed 's/Address: *//g')
-#TODO: replace cf-dev.io with $hostname.ci.van when this script is implemented in CI
+IP=$(nslookup $SCF_DOMAIN | grep answer: -A 2 | grep Address: | sed 's/Address: *//g')
```
mook-as: Alternatively, `host -tA "${SCF_DOMAIN}" | awk '{ print $NF }'`

cf-deploy-teardown/tasks/k8s-ready-state-check.sh
```diff
@@ -33,8 +34,8 @@ status "docker info should show overlay2"
 
 # kube-dns shows 4/4 ready
 
-kube_dns=$(kubectl get pods --all-namespaces | grep -q "kube-dns-")
-[[ $kube_dns == *"4/4 Running"* ]]
+kube_dns=$(kubectl get pods --all-namespaces | grep "kube-dns-")
+[[ $kube_dns =~ 4/4\ *Running ]]
```
mook-as: We should fix the 3/3 vs 4/4 thing.  Something like:
```bash
kubectl get pods --namespace=kube-system --selector k8s-app=kube-dns | grep -E '([0-9])/\1 *Running'
```

prabalsharma: why do I need "--selector k8s-app=kube-dns "

mook-as: It says, "find all pods, in the `kube-system` namespace, with a label `k8s-app` of value `kube-dns`."  Otherwise you'll find other pods in that namespace.

mook-as:
Eh, because I just went for any number.  It was taken from the generic script that just checked the two numbers matched (irrespective of which pod it was).

PR #9, opened 2017-07-05
Provosion CI K8S host with vagrantbox

HartS:
This PR adds a provisioning script for our CI kubernetes hosts, which uses the same libvirt image as our vagrant deployment (without actually using vagrant)

NOTE: The URL for the vagrantbox will need to be updated once the work in SCF is merged

### DIFF
--- a/kube-kvm/README.md
+++ b/kube-kvm/README.md
@@ -41,7 +41,7 @@ should also run on other KVM hosts which meet the following requirements, though
 testing has been done:
 - This script should be used on a KVM host with libvirt installed
 - The qemu user should have access to the path which the disk images will be stored in.
-  This is set by `KUBE_VM_IMAGE_PATH` (defaults to `~/qcow2-disks`)
+  This is set by `KUBE_VM_IMAGE_PATH` (defaults to `/var/lib/libvirt/images/`)
 - The default network defined in `net-default.xml` should be created and started:
   - `virsh net-define net-default.xml`
   - `virsh net-start default`
@@ -65,7 +65,7 @@ Other variables which can be set to modify the behaviour of the deployment scrip
 ```
 KUBE_VM_IMAGE_NAME # The name of the image in s3 to use. Defaults to the latest VM image
 KUBE_VM_IMAGE_PATH # The path to use for storing images, VM disks, and the vm key
-                   # This defaults to $HOME/qcow2-disks
+                   # This defaults to /var/lib/libvirt/images
 KUBE_VM_MEM_GIB    # The memory to allocate to the VM. Defaults to 8GiB
 ```
 
--- a/kube-kvm/README.md
+++ b/kube-kvm/README.md
@@ -1,0 +1,111 @@
+# Provisioning a Kubernetes host on KVM, without vagrant
+
+# Table of Contents
+
+  * [Overview](#overview)
+    * [Why not always use Vagrant?](#why-not-always-use-vagrant)
+  * [How to use this script](#how-to-use-this-script)
+    * [Prerequisites](#prerequisites)
+    * [Configuration](#configuration)
+    * [Deployment](#deployment)
+  * [How to use the VM](#how-to-use-the-vm)
+    * [Installing Dev Tools](#installing-dev-tools)
+
+
+# Overview
+
+This script provides a general-purpose Kubernetes host, with kube-dns and helm deployed,
+which is hospitable for SCF deployment
+
+## Why Not Always Use Vagrant?
+
+To be clear, the KVM image used by this deployment method is *packaged as a vagrant box* by
+the packer vagrant post-provisioner. Even though the input is a vagrant box, using vagrant
+to deploy it introduces a limitation of only being able to deploy one VM at a time on a
+given KVM host. In order to fully utilize shared KVM hosts for development and testing, as
+well as provide Kubernetes host pools for our CI systems, it is advantageous to have a way
+to deploy a VM from the generated KVM image without using Vagrant.
+
+It's important to be aware that the base image built by packer is intended to be a generic
+kube host, and doesn't have direnv, fissile, or stampy. The script in scf
+`bin/dev/install_tools.sh` can be run to install these tools, necessary for building
+releases, images, and charts used for deploying SCF in development. SCF can also be
+deployed from our chart bundle distributions, which use images on our docker hub.
+
+# How to Use This Script
+
+## Prerequisites
+
+This script is intended to be run on a SLES box which has been installed as a KVM host. It
+should also run on other KVM hosts which meet the following requirements, though limited
+testing has been done:
+- The `libvirt` user should be a member of the `root` group: `usermod -aG libvirt root`
+- The default network defined in `net-default.xml` should be created and started:
+  - `virsh net-define net-default.xml`
+  - `virsh net-start default`
+- The ntp service should be installed, synced, and started
+  - `sudo systemctl stop ntpd` (if already started)
+  - `sudo ntpdate time.nist.gov`
+  - `sudo systemctl enable ntpd`
+  - `sudo systemctl start ntpd`
+- The KVM host machine should have internet access through a Linux host bridge device named
+  `br0`. This is the default configuration when installing SLES as a KVM host. If using
+  Wicked as your network manager, you can follow the instructions in step 4 of
+  https://github.com/suse/scf#deploying to configure this bridge interface.
+
+## Configuration
+
+The `provision-kube-host-kvm.sh` script takes one argument, the name to use for the
+deployed Kubernetes host VM. This may alternatively be configured by setting the
+`KUBE_VM_NAME` environment variable.
+
+#### WARNING:
+```
+If a VM with the name passed via the name argument already exists, it will be deleted!
+```
+
+Other variables which can be set to modify the behaviour of the deployment scripts are:
+
+```
+KUBE_VM_IMAGE_NAME # The name of the image in s3 to use. Defaults to the latest VM image
+KUBE_VM_IMAGE_PATH # The path to use for storing images, VM disks, and the vm key
+                   # This defaults to $HOME/qcow2-disks
+KUBE_VM_MEM_GIB    # The memory to allocate to the VM. Defaults to 8GiB
+```
+
+## Deployment
+
+To deploy a new VM, run `./provision-kube-host-kvm.sh <vm-name>`. See the above Warning
+
+# How to Use the VM
+
+Once the deployment of your VM is finished, the provisioning script will output information
+about how to set your kubernetes config to access the VM, which will look like:
+
+```
+kubectl config set-cluster --server=10.9.169.158:8080 my_vm
+kubectl config set-context my_vm --cluster=my_vm
+kubectl config use-context my_vm
+```
+
+You can also access the VM via ssh using the password `vagrant` for `vagrant@$IP`, or the
+default vagrant VM access key which is downloaded as part of the provisioning to
+`"${KUBE_VM_IMAGE_PATH}/vm-key"`
+
+## Installing Dev Tools
+
+When deploying a Kubernetes host with the provisioning script, direnv and the dev tools are
+not installed. Should you wish to prepare the VM so that you can run `make vagrant-prep`,
+you will need to run the following as the `vagrant` user:
+
+```
+mkdir -p /home/vagrant/bin
+wget -O /home/vagrant/bin/direnv --no-verbose \
+  https://github.com/direnv/direnv/releases/download/v2.11.3/direnv.linux-amd64
+chmod a+x /home/vagrant/bin/direnv
+echo 'eval "$(/home/vagrant/bin/direnv hook bash)"' >> ${HOME}/.bashrc
+ln -s /home/vagrant/scf/bin/dev/vagrant-envrc ${HOME}/.envrc
+/home/vagrant/bin/direnv allow ${HOME}
+/home/vagrant/bin/direnv allow ${HOME}/scf
+sudo -E SCF_BIN_DIR=/usr/local/bin HOME=/home/vagrant /home/vagrant/bin/direnv exec /home/vagrant/scf/bin/dev/install_tools.sh
+```
--- a/kube-kvm/README.md
+++ b/kube-kvm/README.md
@@ -43,11 +43,7 @@ testing has been done:
 - The default network defined in `net-default.xml` should be created and started:
   - `virsh net-define net-default.xml`
   - `virsh net-start default`
-- The ntp service should be installed, synced, and started
-  - `sudo systemctl stop ntpd` (if already started)
-  - `sudo ntpdate time.nist.gov`
-  - `sudo systemctl enable ntpd`
-  - `sudo systemctl start ntpd`
+- A system time service (ntp or timedatectl) should be installed, synced, and started
 - The KVM host machine should have internet access through a Linux host bridge device named
   `br0`. This is the default configuration when installing SLES as a KVM host. If using
   Wicked as your network manager, you can follow the instructions in step 4 of
@@ -60,9 +56,7 @@ deployed Kubernetes host VM. This may alternatively be configured by setting the
 `KUBE_VM_NAME` environment variable.
 
 #### WARNING:
-```
-If a VM with the name passed via the name argument already exists, it will be deleted!
-```
+_**If a VM with the name passed via the name argument already exists, it will be deleted!**_
 
 Other variables which can be set to modify the behaviour of the deployment scripts are:
 
@@ -75,7 +69,8 @@ KUBE_VM_MEM_GIB    # The memory to allocate to the VM. Defaults to 8GiB
 
 ## Deployment
 
-To deploy a new VM, run `./provision-kube-host-kvm.sh <vm-name>`. See the above Warning
+To deploy a new VM, run `./provision-kube-host-kvm.sh <vm-name>`. See the above
+[warning](#warning)
 
 # How to Use the VM
 
@@ -83,7 +78,7 @@ Once the deployment of your VM is finished, the provisioning script will output 
 about how to set your kubernetes config to access the VM, which will look like:
 
 ```
-kubectl config set-cluster --server=10.9.169.158:8080 my_vm
+kubectl config set-cluster --server=${IP}:8080 my_vm
 kubectl config set-context my_vm --cluster=my_vm
 kubectl config use-context my_vm
 ```
--- a/kube-kvm/README.md
+++ b/kube-kvm/README.md
@@ -39,7 +39,9 @@ deployed from our chart bundle distributions, which use images on our docker hub
 This script is intended to be run on a SLES box which has been installed as a KVM host. It
 should also run on other KVM hosts which meet the following requirements, though limited
 testing has been done:
-- The `libvirt` user should be a member of the `root` group: `usermod -aG libvirt root`
+- This script should be used on a KVM host with libvirt installed
+- The qemu user should have access to the path which the disk images will be stored in.
+  This is set by `KUBE_VM_IMAGE_PATH` (defaults to `~/qcow2-disks`)
 - The default network defined in `net-default.xml` should be created and started:
   - `virsh net-define net-default.xml`
   - `virsh net-start default`
--- a/kube-kvm/net-default.xml
+++ b/kube-kvm/net-default.xml
@@ -1,0 +1,10 @@
+<network connections='1'>
+  <name>default</name>
+  <forward mode='nat'/>
+  <bridge name='virbr0'/>
+  <ip address='192.168.122.1' netmask='255.255.255.0'>
+    <dhcp>
+      <range start='192.168.122.2' end='192.168.122.254'/>
+    </dhcp>
+  </ip>
+</network>
--- a/kube-kvm/provision-kube-host-kvm.sh
+++ b/kube-kvm/provision-kube-host-kvm.sh
@@ -1,13 +1,14 @@
 #!/usr/bin/env bash
 
 # This script is intended to be run on a SLES box installed as a KVM host.
-# After OS installation, the following commands should also be run to set up directory permissions and the default network (as root):
-# usermod -aG libvirt root
-# virsh net-define net-default.xml
-# virsh net-start default
-# ntpdate time.nist.gov
-# systemctl enable ntpd
-# systemctl start ntpd
+# After OS installation, the following commands should also be run to set up directory
+# permissions and the default network (as root):
+# - usermod -aG libvirt root
+# - virsh net-define net-default.xml
+# - virsh net-start default
+# - ntpdate time.nist.gov
+# - systemctl enable ntpd
+# - systemctl start ntpd
 
 set -o errexit
 
@@ -23,7 +24,7 @@ export OBJECT_STORAGE_YAML_BUCKET=${OBJECT_STORAGE_YAML_BUCKET:-kube-config}
 mkdir -p "$KUBE_VM_IMAGE_PATH"
 cd "$KUBE_VM_IMAGE_PATH"
 
-if [[ -z $KUBE_VM_NAME ]]; then
+if [[ -z "$KUBE_VM_NAME" ]]; then
   echo "A VM name must be provided"
   exit 1
 fi
@@ -80,15 +81,15 @@ unset KUBE_VM_DEFAULT_IP
 # TODO: Clean up variable assignment (timeouts)
 IP_WAIT_TIMEOUT=150
 KUBE_VM_DEFAULT_MAC=$(virsh domiflist "$KUBE_VM_NAME" | grep default | grep -oE '([[:xdigit:]]{2}:){5}[[:xdigit:]]{2}') || true
-if [[ -z $KUBE_VM_DEFAULT_MAC ]]; then
+if [[ -z "$KUBE_VM_DEFAULT_MAC" ]]; then
   echo "No MAC address for default interface on VM $KUBE_VM_NAME. Check output of \`sudo virsh domiflist $KUBE_VM_NAME\`"
   exit 1
 fi
-while [[ -z $KUBE_VM_DEFAULT_IP ]]; do
+while [[ -z "$KUBE_VM_DEFAULT_IP" ]]; do
   echo "Checking if IP lease exists for interface $KUBE_VM_DEFAULT_MAC"
-  KUBE_VM_DEFAULT_IP=$(virsh net-dhcp-leases default | grep "$KUBE_VM_DEFAULT_MAC" | grep -oE '([0-9]{1,3}\.){3}[0-9]{1,3}') || true
-  if [[ -z $KUBE_VM_DEFAULT_IP ]]; then
-    if [[ $IP_WAIT_TIMEOUT -ge 0 ]]; then
+  KUBE_VM_DEFAULT_IP=$(virsh net-dhcp-leases default --mac "$KUBE_VM_DEFAULT_MAC" | grep -oE '([0-9]{1,3}\.){3}[0-9]{1,3}') || true
+  if [[ -z "$KUBE_VM_DEFAULT_IP" ]]; then
+    if [[ "$IP_WAIT_TIMEOUT" -ge 0 ]]; then
       (( IP_WAIT_TIMEOUT -= 15 ))
       echo "No IP assigned for $KUBE_VM_DEFAULT_MAC. Retrying in 15 seconds"
       sleep 15
@@ -101,25 +102,21 @@ done
 echo "Docker VM IP found: $KUBE_VM_DEFAULT_IP. Waiting for sshd to start"
 SSHD_TIMEOUT=60
 while [[ $SSHD_TIMEOUT -ge 0 ]]; do
-  echo '' | nc $KUBE_VM_DEFAULT_IP 22 | grep -i ssh && break
+  nc $KUBE_VM_DEFAULT_IP 22 </dev/null | grep 'SSH-' && break
   sleep 5
   (( SSHD_TIMEOUT -= 5 ))
 done
 ssh-keygen -R $KUBE_VM_DEFAULT_IP
-ssh -o StrictHostKeyChecking=no -Ti "$KUBE_VM_KEY" vagrant@$KUBE_VM_DEFAULT_IP << EOF
-sudo bash -o errexit << EOSU
+ssh -o StrictHostKeyChecking=no -Ti "$KUBE_VM_KEY" vagrant@$KUBE_VM_DEFAULT_IP -- sudo bash -o errexit << EOF
 hostname $KUBE_VM_NAME
 echo $KUBE_VM_NAME > /etc/hostname
 cp /etc/sysconfig/network/ifcfg-eth0 /etc/sysconfig/network/ifcfg-eth1
 echo DHCLIENT_SET_DEFAULT_ROUTE='yes' /etc/sysconfig/network/ifcfg-eth1
 wicked ifup eth1
-EOSU
 EOF
 
-echo "Sleeping 15 seconds for bridged interface to obtain IP"
-sleep 15
-KUBE_VM_BRIDGED_IP=$(ssh -i "$KUBE_VM_KEY" vagrant@$KUBE_VM_DEFAULT_IP '/sbin/ifconfig eth1' | grep -oE 'addr:10\.[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}' | sed 's/addr://g')
-if [[ -z $KUBE_VM_BRIDGED_IP ]]; then
+KUBE_VM_BRIDGED_IP=$(ssh -i "$KUBE_VM_KEY" vagrant@$KUBE_VM_DEFAULT_IP 'ip -4 addr show eth1' | tr / ' ' | awk '/inet/ { print $2 }')
+if [[ -z "$KUBE_VM_BRIDGED_IP" ]]; then
   echo "There was a problem getting the IP of the bridged interface. See output of \`ifconfig eth1\` in VM $KUBE_VM_NAME"
   exit 1
 fi
--- a/kube-kvm/provision-kube-host-kvm.sh
+++ b/kube-kvm/provision-kube-host-kvm.sh
@@ -13,7 +13,7 @@
 set -o errexit
 
 export KUBE_VM_NAME=${KUBE_VM_NAME:-$1}
-export KUBE_VM_IMAGE_NAME=${KUBE_VM_IMAGE_NAME:-scf-libvirt-v2.0.6}
+export KUBE_VM_IMAGE_NAME=${KUBE_VM_IMAGE_NAME:-scf-libvirt-v2.0.7}
 export KUBE_VM_IMAGE_PATH=${KUBE_VM_IMAGE_PATH:-~/qcow2-disks}
 export KUBE_VM_MEM_GIB=${KUBE_VM_MEM_GIB:-8}
 
--- a/kube-kvm/provision-kube-host-kvm.sh
+++ b/kube-kvm/provision-kube-host-kvm.sh
@@ -17,53 +17,54 @@ export KUBE_VM_IMAGE_NAME=${KUBE_VM_IMAGE_NAME:-scf-libvirt-v2.0.7}
 export KUBE_VM_IMAGE_PATH=${KUBE_VM_IMAGE_PATH:-~/qcow2-disks}
 export KUBE_VM_MEM_GIB=${KUBE_VM_MEM_GIB:-8}
 
-mkdir -p "$KUBE_VM_IMAGE_PATH"
-cd "$KUBE_VM_IMAGE_PATH"
-
-if [[ -z "$KUBE_VM_NAME" ]]; then
+if [[ -z "${KUBE_VM_NAME}" ]]; then
   echo "A VM name must be provided"
   exit 1
 fi
 
+mkdir -p "${KUBE_VM_IMAGE_PATH}"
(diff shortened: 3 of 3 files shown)


k8s-kvm/provision_k8s_host_kvm.sh
```diff
@@ -0,0 +1,145 @@
...
+done
+echo "Docker VM IP found: $K8S_VM_DEFAULT_IP. Waiting for sshd to start"
+SSHD_TIMEOUT=60
+while [[ $SSHD_TIMEOUT -ge 0 ]]; do
+  echo '' | nc $K8S_VM_DEFAULT_IP 22 | grep -i ssh && break
+  sleep 5
+  (( SSHD_TIMEOUT -= 5 ))
+done
+ssh-keygen -R $K8S_VM_DEFAULT_IP
+ssh -o StrictHostKeyChecking=no -Ti "$K8S_VM_KEY" vagrant@$K8S_VM_DEFAULT_IP << EOF
```
mook-as: You might be able to get away with
```bash
ssh … vagrant@${K8S_VM_DEFAULT_IP} -- sudo bash -o errexit << EOF
…
EOF
```

k8s-kvm/provision_k8s_host_kvm.sh
```diff
@@ -0,0 +1,145 @@
...
+sudo bash -o errexit << EOSU
+hostname $K8S_VM_NAME
+echo $K8S_VM_NAME > /etc/hostname
```
mook-as: Are we assuming SUSE here? If yes, `yast dns edit hostname="${K8S_VM_NAME}"` might work here.

k8s-kvm/provision_k8s_host_kvm.sh
```diff
@@ -0,0 +1,145 @@
...
+cp /etc/sysconfig/network/ifcfg-eth0 /etc/sysconfig/network/ifcfg-eth1
+echo DHCLIENT_SET_DEFAULT_ROUTE='yes' /etc/sysconfig/network/ifcfg-eth1
+wicked ifup eth1
+EOSU
+EOF
+
+echo "Sleeping 15 seconds for bridged interface to obtain IP"
+sleep 15
```
mook-as: Avoid hard-coded (and non-retrying) timeouts, poll for success instead.

k8s-kvm/provision_k8s_host_kvm.sh
```diff
@@ -0,0 +1,145 @@
...
+K8S_VM_BRIDGED_IP=$(ssh -i "$K8S_VM_KEY" vagrant@$K8S_VM_DEFAULT_IP '/sbin/ifconfig eth1' | grep -oE 'addr:10\.[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}' | sed 's/addr://g')
```
mook-as: `ifconfig` is deprecated; consider using `ip -4 addr show eth1 | tr / ' ' | awk '/inet/ { print $2 }'` instead.

k8s-kvm/provision_k8s_host_kvm.sh
```diff
@@ -0,0 +1,145 @@
...
+  exit 1
+fi
+
+ssh-keygen -R $K8S_VM_BRIDGED_IP
+ssh -o StrictHostKeyChecking=no -Ti "$K8S_VM_KEY" vagrant@$K8S_VM_BRIDGED_IP  << EOF
+git clone https://github.com/suse/scf
+cd scf
+sudo bash -o errexit << EOSU
+export SCF_BIN_DIR=/usr/local/bin
+./bin/common/install_tools.sh
```
mook-as: Does this need to run as root?

prep-yml-bundle/rename-kube-domain.sh
```diff
@@ -0,0 +1,81 @@
+#!/bin/bash
+
+function usage() {
+  echo "Usage: ./rename-kube-domain.sh DOMAIN path/to/yamls path/to/scf path/to/uaa-fissile-release"
```
mook-as: You're going to need to expand on this to say what those arguments mean…

prep-yml-bundle/rename-kube-domain.sh
```diff
@@ -0,0 +1,81 @@
...
+}
+
+# This script takes a new domain name and a path containing a yml bundle of all cf and uaa ymls, recursively copies it, and performs all necessary domain and cert generations from the uaa-fissile-release and scf repos (also passed as arguments) and substitutions in the copied yml bundle
```
mook-as: Please line-wrap as appropriate.

mook-as: Is it just me, or is this screaming "use the helm version"?

prep-yml-bundle/rename-kube-domain.sh
```diff
@@ -0,0 +1,81 @@
...
+  exit 1
+fi
+
+cp -r "$BUNDLE_PATH" $WORKDIR
+
+sed -i "s/DOMAIN=.*/DOMAIN=$DOMAIN/g" "$UAA_PATH/env/defaults.env"
+"$UAA_PATH/generate-certs.sh" $WORKDIR/uaa/certs.env
+(
+  cd $UAA_PATH
+  git checkout env/defaults.env
```
mook-as: This should be working off a dist zip file, I think, rather than a checkout.

prep-yml-bundle/replace-vars.rb
```diff
@@ -0,0 +1,32 @@
+#!/usr/bin/ruby
+
+# Usage ./replace_vars.rb COMMA,SEPARATED,VAR_NAMES path/to.yml [path/to.yml ... ]
+
+# This script will walk through all yaml files provided, and for any object that has a key 'name' that matches
+# the name of a variable to replace, it will replace the value of 'value' in the yaml with the value of the
+# variable from the environment. Requires Ruby >= 2.1.0
```
mook-as: Line wrap problems (a bit too wide).

Reading the code, this actually only changes `spec/template/spec/containers/env/*`, not what is described at all.  The description made me think it would do so regardless of where in the hierarchy the key is.

k8s-kvm/provision_k8s_host_kvm.sh
```diff
@@ -0,0 +1,145 @@
...
+# usermod -aG libvirt root
+# virsh net-define net-default.xml
+# virsh net-start default
+# ntpdate time.nist.gov
+# systemctl enable ntpd
+# systemctl start ntpd
+
+set -o errexit
+
+export K8S_VM_NAME=${K8S_VM_NAME:-$1}
```
HartS: I replaced all ~60 K8S prefixes

k8s-kvm/provision_k8s_host_kvm.sh
```diff
@@ -0,0 +1,145 @@
...
+export K8S_VM_MEM_GIB=${K8S_VM_MEM_GIB:-8}
+
+# Used for displaying external setup information at end of provisioning
+export OBJECT_STORAGE_BASE_URL=${OBJECT_STORAGE_BASE_URL:-\${OBJECT_STORAGE_BASE_URL}}
+export OBJECT_STORAGE_YAML_BUCKET=${OBJECT_STORAGE_YAML_BUCKET:-kube-config}
+
+mkdir -p "$K8S_VM_IMAGE_PATH"
+cd "$K8S_VM_IMAGE_PATH"
+
+if [[ -z $K8S_VM_NAME ]]; then
```
HartS: Quotes *shouldn't* be needed when using double square brackets (except for maybe with string comparison?), but I'll go ahead and quote anyway for aesthetic purposes

k8s-kvm/provision_k8s_host_kvm.sh
```diff
@@ -0,0 +1,145 @@
...
+  --os-variant opensuse42.2 \
+  --wait 0 \
+  --hvm
+
+echo "Sleeping 30 seconds to allow VM to start and obtain IP"
+sleep 30 
+unset K8S_VM_DEFAULT_IP
+# TODO: Clean up variable assignment (timeouts)
+IP_WAIT_TIMEOUT=150
+K8S_VM_DEFAULT_MAC=$(virsh domiflist "$K8S_VM_NAME" | grep default | grep -oE '([[:xdigit:]]{2}:){5}[[:xdigit:]]{2}') || true
```
HartS: I generally prefer using regexes that describe what I'm selecting to relying on the position of a column (in case competing versions have a different order/number of columns)

k8s-kvm/provision_k8s_host_kvm.sh
HartS: Yes; currently it's putting the binaries into `/usr/local/bin`

k8s-kvm/provision_k8s_host_kvm.sh
```diff
@@ -0,0 +1,145 @@
...
+
+if [[ $K8S_VM_MEM_GIB -gt $(( $(free -g | grep [0-9] | head -1 | awk '{print $2}') - 2 )) ]]; then
+  echo K8S_VM_MEM_GIB value $K8S_VM_MEM_GIB exceeds system resources
+  exit 1
+fi
+
+unset RUNNING_VM_IMAGE
+if virsh domstate $K8S_VM_NAME 2>/dev/null; then
+  echo "Deleting existing domain $K8S_VM_NAME"
+  RUNNING_VM_IMAGE=$(virsh dumpxml $K8S_VM_NAME | grep 'source file' | grep -o "'.*'" | tr -d "'")
```
HartS: Thanks for that pointer; I've never used xmlstarlet. I have concerns about whether it's safe to assume that it's installed though.. it doesn't seem to be installed in some SUSE-based systems by default

k8s-kvm/provision_k8s_host_kvm.sh
```diff
@@ -0,0 +1,145 @@
...
+  --disk "$K8S_VM_IMAGE_NAME-$K8S_VM_NAME.qcow2" \
+  --import \
+  --network network=default \
+  --network bridge=br0 \
+  --os-variant opensuse42.2 \
+  --wait 0 \
+  --hvm
+
+echo "Sleeping 30 seconds to allow VM to start and obtain IP"
+sleep 30
```
HartS: I can definitely loop over a shorter period, until the success state, but I believe a timeout is still in order, since I don't know what kind of errors I might encounter if something goes wrong.

k8s-kvm/provision_k8s_host_kvm.sh
HartS: This is intended to run on any kvm host which satisfies the requirements. Practically speaking, this will probably be a SLE kvm host every time, but is there a reason to prefer the SUSE-specific hostname rename instead of a more general one?

k8s-kvm/provision_k8s_host_kvm.sh
HartS: I don't think a sleep is strictly needed here at all; I added it to avoid potential issues where the `wicked ifup` command has completed. I'm going to try removing this sleep altogether, and if that fails I'll have an idea of what the condition to sleep and retry should be.

prep-yml-bundle/rename-kube-domain.sh
HartS: :man_facepalming: from here down can be ignored. These files should have been deleted now that we're using helm.

k8s-kvm/provision_k8s_host_kvm.sh
mook-as: Ah, didn't realize you wanted it that generic. The reason I ask for yast is because it seems to tickle more places and updates the hostname in the prompt faster

k8s-kvm/provision_k8s_host_kvm.sh
HartS: This seems to work fine without the sleep here.

k8s-kvm/provision_k8s_host_kvm.sh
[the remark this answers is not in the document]
HartS: Requesting some additional feedback here

k8s-kvm/provision_k8s_host_kvm.sh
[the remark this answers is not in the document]
mook-as: Ah, right, `[[` might have fixed things. Thanks.

k8s-kvm/provision_k8s_host_kvm.sh
```diff
@@ -0,0 +1,145 @@
...
+fi
+
+# Fetch VM access key if it doesn't exist
+if [[ ! -f vm-key ]]; then
+  curl -sL -o "vm-key" https://raw.githubusercontent.com/mitchellh/vagrant/v1.9.6/keys/vagrant
+  chmod 400 vm-key
+fi
+K8S_VM_KEY=${K8S_VM_KEY:-$PWD/vm-key} 
+
+if [[ $K8S_VM_MEM_GIB -gt $(( $(free -g | grep [0-9] | head -1 | awk '{print $2}') - 2 )) ]]; then
```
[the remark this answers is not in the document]
mook-as: Confused by the 👍 — it isn't a critical change, but you seemed to agree, but didn't change anything 😄  (Will be happy to merge as-is, if nothing else blocks.)

k8s-kvm/provision_k8s_host_kvm.sh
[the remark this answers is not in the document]
mook-as: Yeah, the downside of that is availability. Too bad there isn't a bash-native XML parser.

… on second thought, maybe it's better that there isn't a bash-native XML parser.

k8s-kvm/provision_k8s_host_kvm.sh
```diff
@@ -0,0 +1,145 @@
...
+  if [[ -z $RUNNING_VM_IMAGE ]]; then
+    echo "Disk for domain $K8S_VM_NAME could not be determined. Please delete this domain manually"
+    exit 1
+  fi
+  virsh destroy "$K8S_VM_NAME"
+  virsh undefine "$K8S_VM_NAME"
+  rm -f "$RUNNING_VM_IMAGE"
+fi
+
+cp "$K8S_VM_IMAGE_NAME.qcow2" "$K8S_VM_IMAGE_NAME-$K8S_VM_NAME.qcow2"
```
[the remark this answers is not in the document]
mook-as: This seems to be missing.  It's the difference between ~instant and ~minutes, so please have it 😁

k8s-kvm/provision_k8s_host_kvm.sh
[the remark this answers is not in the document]
mook-as: Sure, having a (long) timeout is fine. It's just that I don't trust anything to complete within 30s.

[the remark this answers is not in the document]
HartS: Does 1 minute seem reasonable with 15 second retry intervals?

k8s-kvm/provision_k8s_host_kvm.sh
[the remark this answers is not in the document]
HartS: This tends to take ~15s for me IIRC, though I don't doubt it could take minutes depending on your disk/FS. Even with `reflink=auto`.. isn't the actual filesystem copy going to occur as soon as the VM is started? If I understand correctly, that just means we're postponing this overhead for one command. With the copy already behind us, I haven't seen it take more than 30 seconds to get the IP from the virtual network (though I don't doubt it can happen).. see below where I suggest a 1 minute timeout.. maybe that needs to be increased as well? The problem is, I have no way of knowing what a reasonable timeout there is. Especially if the actual on-disk copy is happening at that point, couldn't that take upwards of 5 minutes for users with slow disks?

k8s-kvm/provision_k8s_host_kvm.sh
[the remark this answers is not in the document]
HartS: See comment on `--reflink=auto` above

k8s-kvm/provision_k8s_host_kvm.sh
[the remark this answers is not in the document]
mook-as: You'd normally not have to write every cluster of the image (most of it is read-only data, things like binaries). So only changes would need to be copied.

k8s-kvm/provision_k8s_host_kvm.sh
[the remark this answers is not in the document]
HartS: Er.. the timeout *is* already 2 minutes and 30s, it just waits 30s to make the initial attempt to get the IP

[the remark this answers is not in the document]
mook-as: I'd bump the 1 minute timeout to something much longer (say, 5 or 10), since we never expect to hit it anyway, it's just there in case we try it and walk away.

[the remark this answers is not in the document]
HartS: OK.. it waits 30s to get the MAC address.. that should be available as soon as the VM is started. Retrying here is reasonable.. the script then waits up to 2:30 to get the IP.. I think that's reasonable, but as mentioned, if we don't actually make the copy until this step, I'm concerned it could take longer.

kube-kvm/provision-kube-host-kvm.sh
```diff
@@ -0,0 +1,147 @@
...
+fi
+
+# Fetch VM access key if it doesn't exist
+if [[ ! -f vm-key ]]; then
+  curl -sL -o "vm-key" https://raw.githubusercontent.com/mitchellh/vagrant/v1.9.6/keys/vagrant
+  chmod 400 vm-key
+fi
+KUBE_VM_KEY=${KUBE_VM_KEY:-$PWD/vm-key}
+
+if [[ $KUBE_VM_MEM_GIB -gt $(free -g | awk 'NR == 2 { print $2 - 2 }') ]]; then
```
mook-as: What is the hard coded `2` you are subtracting for?

kube-kvm/provision-kube-host-kvm.sh
```diff
@@ -0,0 +1,147 @@
...
+  echo KUBE_VM_MEM_GIB value $KUBE_VM_MEM_GIB exceeds system resources
+  exit 1
+fi
+
+unset RUNNING_VM_IMAGE
+if virsh domstate $KUBE_VM_NAME 2>/dev/null; then
+  echo "Deleting existing domain $KUBE_VM_NAME"
+  RUNNING_VM_IMAGE=$(virsh dumpxml $KUBE_VM_NAME | grep 'source file' | grep -o "'.*'" | tr -d "'")
+  if [[ -z $RUNNING_VM_IMAGE ]]; then
```
mook-as: Please quote things that might be empty.  Ideally quote everything, but especially things that might be empty.

kube-kvm/provision-kube-host-kvm.sh
```diff
@@ -0,0 +1,147 @@
...
+    echo "Disk for domain $KUBE_VM_NAME could not be determined. Please delete this domain manually"
+    exit 1
+  fi
+  virsh destroy "$KUBE_VM_NAME"
+  virsh undefine "$KUBE_VM_NAME"
+  rm -f "$RUNNING_VM_IMAGE"
+fi
+
+cp --reflink=auto "$KUBE_VM_IMAGE_NAME.qcow2" "$KUBE_VM_IMAGE_NAME-$KUBE_VM_NAME.qcow2"
```
mook-as: Actually, why do we need a copy again? (As opposed to a `qemu-img create -b`)

kube-kvm/provision-kube-host-kvm.sh
```diff
@@ -0,0 +1,147 @@
...
+  --disk "$KUBE_VM_IMAGE_NAME-$KUBE_VM_NAME.qcow2" \
+  --import \
+  --network network=default \
+  --network bridge=br0 \
+  --os-variant opensuse42.2 \
+  --wait 0 \
+  --hvm
+
+echo "Sleeping 30 seconds to allow VM to start and obtain IP"
+sleep 30
```
mook-as: We don't really need the sleep outside the mac-fetching loop do we?

kube-kvm/provision-kube-host-kvm.sh
```diff
@@ -0,0 +1,147 @@
...
+    else
+      echo "No IP assigned for $KUBE_VM_DEFAULT_MAC after 10 minutes. Exiting."
+      exit 1
+    fi
+  fi
+done
+echo "Docker VM IP found: $KUBE_VM_DEFAULT_IP. Waiting for sshd to start"
+SSHD_TIMEOUT=60
+while [[ $SSHD_TIMEOUT -ge 0 ]]; do
+  nc $KUBE_VM_DEFAULT_IP 22 </dev/null | grep 'SSH-' && break
```
mook-as: Since you're in bash, you can use `head -c4 </dev/tcp/${KUBE_VM_DEFAULT_IP}/22`… but `nc` is more sane. Let's stick with that.

kube-kvm/provision-kube-host-kvm.sh
```diff
@@ -0,0 +1,147 @@
...
+  sleep 5
+  (( SSHD_TIMEOUT -= 5 ))
+done
+ssh-keygen -R $KUBE_VM_DEFAULT_IP
+ssh -o StrictHostKeyChecking=no -Ti "$KUBE_VM_KEY" vagrant@$KUBE_VM_DEFAULT_IP -- sudo bash -o errexit << EOF
+hostname $KUBE_VM_NAME
+echo $KUBE_VM_NAME > /etc/hostname
+cp /etc/sysconfig/network/ifcfg-eth0 /etc/sysconfig/network/ifcfg-eth1
+echo DHCLIENT_SET_DEFAULT_ROUTE='yes' /etc/sysconfig/network/ifcfg-eth1
```
mook-as: Did you mean to pipe that line into the file? Right now you're just echoing two words.

kube-kvm/provision-kube-host-kvm.sh
```diff
@@ -0,0 +1,147 @@
...
+wicked ifup eth1
```
mook-as: I… don't understand. `wicked` is, as far as I know, SUSE-specific…

kube-kvm/provision-kube-host-kvm.sh
```diff
@@ -0,0 +1,147 @@
...
+  exit 1
+fi
+
+ssh-keygen -R $KUBE_VM_BRIDGED_IP
+ssh -o StrictHostKeyChecking=no -Ti "$KUBE_VM_KEY" vagrant@$KUBE_VM_BRIDGED_IP  << EOF
+git clone https://github.com/suse/scf
+cd scf
+sudo bash -o errexit << EOSU
+export SCF_BIN_DIR=/usr/local/bin
+./bin/common/install_tools.sh
```
mook-as: Why does the VM need tools again? I thought it was meant to be a kube-only VM (and not even for actually running `kubectl` on; rather, just running the kube server part)

kube-kvm/net-default.xml
```diff
@@ -0,0 +1,10 @@
+<network connections='1'>
+  <name>default</name>
```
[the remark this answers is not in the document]
HartS: IIRC this isn't created in all installations. Best to leave it here for reference I think.

kube-kvm/provision-kube-host-kvm.sh
```diff
@@ -0,0 +1,147 @@
...
+# - virsh net-define net-default.xml
+# - virsh net-start default
+# - ntpdate time.nist.gov
+# - systemctl enable ntpd
+# - systemctl start ntpd
+
+set -o errexit
+
+export KUBE_VM_NAME=${KUBE_VM_NAME:-$1}
+export KUBE_VM_IMAGE_NAME=${KUBE_VM_IMAGE_NAME:-scf-libvirt-v2.0.7}
```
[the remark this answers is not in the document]
HartS: That requires cloning the SCF directory.. if we change directory permissions on https://cf-opensusefs2.s3.amazonaws.com/vagrant/ we could parse it out of the output there?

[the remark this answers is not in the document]
mook-as: Ah, I didn't realize you expected people to run this without a full checkout. Nevermind, then, if that's the case. (How do they get the script otherwise?)

kube-kvm/README.md
```diff
@@ -0,0 +1,111 @@
...
+
+This script is intended to be run on a SLES box which has been installed as a KVM host. It
+should also run on other KVM hosts which meet the following requirements, though limited
+testing has been done:
+- The `libvirt` user should be a member of the `root` group: `usermod -aG libvirt root`
+- The default network defined in `net-default.xml` should be created and started:
+  - `virsh net-define net-default.xml`
+  - `virsh net-start default`
+- The ntp service should be installed, synced, and started
+  - `sudo systemctl stop ntpd` (if already started)
```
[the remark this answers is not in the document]
HartS: I'm not sure I understand why, but with ntpd running, `ntpdate time.nist.gov` exits, saying `the NTP socket is in use` (but adding `-u` appears to work around this.

Looking into it, I just noticed ntpdate may be deprecated with systemd, in favor of `timedatectl`, so I'm going to remove the mentions of ntp/ntpd altogether and just mention that the clock time must be synced.

kube-kvm/README.md
```diff
@@ -0,0 +1,111 @@
...
+```
+If a VM with the name passed via the name argument already exists, it will be deleted!
+```
+
+Other variables which can be set to modify the behaviour of the deployment scripts are:
+
+```
+KUBE_VM_IMAGE_NAME # The name of the image in s3 to use. Defaults to the latest VM image
+KUBE_VM_IMAGE_PATH # The path to use for storing images, VM disks, and the vm key
+                   # This defaults to $HOME/qcow2-disks
```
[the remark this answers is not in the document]
HartS: Preferably we'd keep the original images around.. Also, if we put the disk data in a temp directory, VMs would break after a restart

kube-kvm/provision-kube-host-kvm.sh
HartS: It's pretty arbitrary, but since we might want to allow overprovisioning, I'm only checking that the user isn't trying to provision a VM with as much memory as the host machine. I thought 2 GB to manage the host OS and KVM overhead would be reasonable, but let me know if you have other ideas.

mook-as: I'm fine with it, as long as we have a comment in the script about that.  No better ideas here.

kube-kvm/provision-kube-host-kvm.sh
HartS: :man_facepalming:

kube-kvm/provision-kube-host-kvm.sh
HartS: The VM should be a SUSE VM? But the host machine may not be

HartS: Ah.. I see what you were saying about setting the hostname

kube-kvm/provision-kube-host-kvm.sh
HartS: The image itself is for a kube-only VM.. installing common tools installs helm, brings in a newer version of kubectl (IIRC), and also adds 'k' and 'cf'. The intention of this script is to provide a base to deploy SCF to, with the assumption that things like bosh release fetching and compilation, and fissile image creation, don't need to happen on it.

mook-as: Right, but the _server_ has no need of `kubectl`, `k`, nor `cf`. `helm` is only useful insofar as installing `tiller` (which can be done on the machine the script is running from).

Running `install_tools.sh` on the server means cloning all of SCF 😞

HartS: True, it doesn't *need* `kubectl`, `k`, and `cf`, but I'd argue they will be useful when debugging when things go wrong in CI (kind of like how the SCF images don't need sshd, less, vi, sed, grep), and reusing `install_tools.sh` ensures we have the same helm version across deployments

Cloning SCF without submodules is fairly quick (it's about 38 MB). I could also cherry-pick the versions.sh and `common/install_tools.sh` to avoid the need to clone it at all.

kube-kvm/provision-kube-host-kvm.sh
HartS:
```
sharedkvmhost1:~/qcow2-disks # qemu-img create -b scf-libvirt-v2.0.7.qcow2  scf-libvirt-v2.0.7-test2.qcow2 
qemu-img: scf-libvirt-v2.0.7-test2.qcow2: Backing file not supported for file format 'raw'
```

mook-as: Well, you'd need to actually create a non-raw image, not just call a raw image `.qcow2` :)
`qemu-img create -f qcow2 -b …`

kube-kvm/provision-kube-host-kvm.sh
HartS: > Are we assuming SUSE here? If yes, yast dns edit hostname="${K8S_VM_NAME}" might work here.

So apparently, even without making any of my own changes, this command doesn't work for me, because both the default `/etc/sysconfig/network/ifcfg-eth0` and `/etc/sysconfig/network/dhcp` files contain the line `DHCLIENT_SET_HOSTNAME='yes'`

I think if I removed this line from both of these files and add it to ifcfg-eth1, I may be able to set the hostname with yast noninteractively, but I'm not clear on what wicked does with this option in the various config.

For simplicity, I'll stick with setting `hostname` without yast for now.

kube-kvm/net-default.xml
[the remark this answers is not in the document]
mook-as: That's fine, but it means the [instructions to create the network](https://github.com/SUSE/cf-ci/pull/9/files#diff-c6d4f3779893640650e8b59701ac4134R44) will need to be conditional.

kube-kvm/provision-kube-host-kvm.sh
```diff
@@ -0,0 +1,148 @@
...
+
+# Fetch VM access key if it doesn't exist
+KUBE_VM_KEY=${KUBE_VM_KEY:-$PWD/vm-key}
+if [[ ! -f "${KUBE_VM_KEY}" ]] && [[ "${KUBE_VM_KEY}" == "$PWD/vm-key" ]]; then
+  curl -sL -o "{KUBE_VM_KEY}" https://raw.githubusercontent.com/mitchellh/vagrant/v1.9.6/keys/vagrant
+  chmod 400 vm-key
+fi
+
+# Check that provisioning the VM would leave host machine with at least 2 GiB for OS/KVM overhead
+if [[ ${KUBE_VM_MEM_GIB} -gt $(free -g | awk 'NR == 2 { print $2 - 2 }') ]]; then
```
mook-as: If you prefer being explicit you'll want `free -g | awk '/Mem:/ { print $2 - 2 }'`

kube-kvm/provision-kube-host-kvm.sh
```diff
@@ -0,0 +1,148 @@
...
+    else
+      echo "No IP assigned for ${KUBE_VM_DEFAULT_MAC} after 10 minutes. Exiting."
+      exit 1
+    fi
+  fi
+done
+echo "Docker VM IP found: ${KUBE_VM_DEFAULT_IP}. Waiting for sshd to start"
+SSHD_TIMEOUT=60
+while [[ ${SSHD_TIMEOUT} -ge 0 ]]; do
+  nc "${KUBE_VM_DEFAULT_IP}" 22 </dev/null | grep 'SSH-' && break
```
mook-as: `grep --quiet`?

kube-kvm/provision-kube-host-kvm.sh
mook-as: I'm confused. You don't expect people to clone scf to get the shared version, but expect cloning it to be quick here?

I'd just go with, let the user do it if they want to, it's already in the readme.

HartS: Users don't need to clone scf to get the version if the version is kept current :) The common tool installation isn't in the README (yet); I can add it there, and remove it from default installation.. I still think this will result in more time spent picking in these tools manually, installing helm manually, and typing out full kubectl commands because 'k' isn't around

kube-kvm/provision-kube-host-kvm.sh
```diff
@@ -0,0 +1,148 @@
...
+# Fetch base image if it doesn't exist
+if [[ ! -f "${KUBE_VM_IMAGE_NAME}.qcow2" ]]; then
+  echo "Fetching base image:"
+  curl -L "https://cf-opensusefs2.s3.amazonaws.com/vagrant/${KUBE_VM_IMAGE_NAME}.box" | tar -xz box.img
+  mv box.img "${KUBE_VM_IMAGE_NAME}.qcow2"
+fi
+
+# Fetch VM access key if it doesn't exist
+KUBE_VM_KEY=${KUBE_VM_KEY:-$PWD/vm-key}
+if [[ ! -f "${KUBE_VM_KEY}" ]] && [[ "${KUBE_VM_KEY}" == "$PWD/vm-key" ]]; then
```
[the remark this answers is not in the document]
HartS: You mean to place both conditions into one bash conditional? To me, that makes it less readable.

kube-kvm/README.md
```diff
@@ -0,0 +1,111 @@
...
+deployed from our chart bundle distributions, which use images on our docker hub.
+
+# How to Use This Script
+
+## Prerequisites
+
+This script is intended to be run on a SLES box which has been installed as a KVM host. It
+should also run on other KVM hosts which meet the following requirements, though limited
+testing has been done:
+- The `libvirt` user should be a member of the `root` group: `usermod -aG libvirt root`
```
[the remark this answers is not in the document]
HartS: So I've noticed that if I don't `usermod -aG root qemu`, VMs in `/root` can't be started.. I'll update the instructions to mention that the qemu user needs to have access to whichever directory the images are going in, and also `chown qemu:qemu $IMAGE` after the `qemu-img create -f` command

kube-kvm/provision-kube-host-kvm.sh
```diff
@@ -0,0 +1,148 @@
...
+# - virsh net-start default
+# - ntpdate time.nist.gov
+# - systemctl enable ntpd
+# - systemctl start ntpd
+
+set -o errexit
+
+export KUBE_VM_NAME=${KUBE_VM_NAME:-$1}
+export KUBE_VM_IMAGE_NAME=${KUBE_VM_IMAGE_NAME:-scf-libvirt-v2.0.7}
+export KUBE_VM_IMAGE_PATH=${KUBE_VM_IMAGE_PATH:-~/qcow2-disks}
```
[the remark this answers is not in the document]
HartS: Two reasons I put this in a separate directory, though I don't necessarily feel strongly that it needs to be:

- We need a place to put the default vagrant VM key. I didn't like the idea of adding files to `~/.ssh`, or putting this in `/var/lib/libvirt/images`
- Placing this in a separate directory (effectively a 'pool' in libvirt terms) keeps the base images as well as the provisioned VM disk images separated. However, the user can set this to the same path as their other libvirt images

I'm not opposed to making the results of that calculation the default for VM_IMAGE_PATH though.

kube-kvm/README.md
```diff
@@ -0,0 +1,106 @@
...
+
+This script provides a general-purpose Kubernetes host, with kube-dns and helm deployed,
+which is hospitable for SCF deployment
+
+## Why Not Always Use Vagrant?
+
+To be clear, the KVM image used by this deployment method is *packaged as a vagrant box* by
+the packer vagrant post-provisioner. Even though the input is a vagrant box, using vagrant
+to deploy it introduces a limitation of only being able to deploy one VM at a time on a
+given KVM host. In order to fully utilize shared KVM hosts for development and testing, as
```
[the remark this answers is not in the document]
HartS: Added a note about this to the requirements in the README

HartS:
Regarding removing `./bin/common/install_tools.sh` from the provisioning script (I can't respond to that comment for some reason)

Cloning happens in the VM to get the provisioning script, but it's necessary, and adds roughly 5 seconds. I wonder if it would make sense to make `scf` a submodule instead, then we could parse the latest version from the Vagrantfile, and also scp `common/install_tools.sh` into the VM. This prevents us from having to clone inside the VM, though bumping and updating the submodule isn't less trivial than updating the .box version reference manually.

kube-kvm/README.md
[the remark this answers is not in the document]
mook-as: Or, we can do the way more sensible thing, and just dump the image in the normal libvirt image directory.  Hence https://github.com/SUSE/cf-ci/pull/9#discussion_r136180155 .

mook-as:
Oh right, this is cf-ci, not scf. Nevermind then, and we can drop the version thing.

kube-kvm/provision-kube-host-kvm.sh
mook-as: Right, the problem is that I totally forgot this is the cf-ci repo, not the scf repo. So it makes sense.

mook-as:
I believe the only issue left is where to put the images :)

kube-kvm/provision-kube-host-kvm.sh
```diff
@@ -14,7 +14,7 @@ set -o errexit
 
 export KUBE_VM_NAME=${1:-${KUBE_VM_NAME}}
 export KUBE_VM_IMAGE_NAME=${KUBE_VM_IMAGE_NAME:-scf-libvirt-v2.0.7}
-export KUBE_VM_IMAGE_PATH=${KUBE_VM_IMAGE_PATH:-~/qcow2-disks}
+export KUBE_VM_IMAGE_PATH=${KUBE_VM_IMAGE_PATH:-/var/lib/libvirt/images}
```
mook-as: I think I could preferred more magic, but this is fine.

PR #12, opened 2017-07-13
Split scf system checks into categories for different types of nodes

andreas-kupries:
In reverse order, top committed last
* Reworked check for storageclass to simply count instead of requiring a specific name. Requested by @aarondl 
*   Added helper for category testing, and annotated all checks with their categories.
*   Added usage information, option processing to invoke it, and category setup.
*   Separated colorization (`red`, `green`) from status types (`verified`, `trouble`).
     Replaced weird `awk`/`print` construction with bash `printf`.

### DIFF
--- a/cf-deploy-teardown/tasks/k8s-ready-state-check.sh
+++ b/cf-deploy-teardown/tasks/k8s-ready-state-check.sh
@@ -1,5 +1,29 @@
 #!/bin/bash
 
+function usage() {
+    cat <<EOF
+Usage: $(basename "${0}") ?options? ?category?
+
+  -h: Displays this help message
+
+  Supported categories: all, api, kube, node
+  Defaults to: all
+EOF
+}
+
+while getopts "h" opt; do
+    case $opt in
+	h)
+	    usage
+	    exit
+	    ;;
+    esac
+done
+
+shift $((OPTIND-1))
+
+category="${1:-all}"
+
 #Script to determine is the K8s host is "ready" for cf deployment
 FAILED=0
 SCF_DOMAIN=${SCF_DOMAIN:-cf-dev.io}
@@ -28,6 +52,9 @@ function status() {
 	FAILED=1
     fi
 }
+
+
+echo "Testing $(green $category)"
 
 # cgroup memory & swap accounting in /proc/cmdline
 grep -wq "cgroup_enable=memory" /proc/cmdline
--- a/cf-deploy-teardown/tasks/k8s-ready-state-check.sh
+++ b/cf-deploy-teardown/tasks/k8s-ready-state-check.sh
@@ -53,56 +53,74 @@ function status() {
     fi
 }
 
+function having_category() {
+    case "$@" in
+	*${category}*)
+	    return 0
+	    ;;
+    esac
+    return 1
+}
 
 echo "Testing $(green $category)"
 
 # cgroup memory & swap accounting in /proc/cmdline
-grep -wq "cgroup_enable=memory" /proc/cmdline
-status "cgroup_enable memory"
+if having_category all node ; then
+    grep -wq "cgroup_enable=memory" /proc/cmdline
+    status "cgroup_enable memory"
 
-grep -wq "swapaccount=1" /proc/cmdline
-status "swapaccount enable"
+    grep -wq "swapaccount=1" /proc/cmdline
+    status "swapaccount enable"
+fi
 
 # docker info should show overlay2
-
-docker info 2> /dev/null | grep -wq "Storage Driver: overlay2"
-status "docker info should show overlay2"
+if having_category all api kube node ; then
+    docker info 2> /dev/null | grep -wq "Storage Driver: overlay2"
+    status "docker info should show overlay2"
+fi
 
 # kube-dns shows 4/4 ready
-
-kubectl get pods --namespace=kube-system --selector k8s-app=kube-dns | grep -Eq '([0-9])/\1 *Running'
-status "kube-dns should shows 4/4 ready"
+if having_category all kube ; then
+    kubectl get pods --namespace=kube-system --selector k8s-app=kube-dns | grep -Eq '([0-9])/\1 *Running'
+    status "kube-dns should shows 4/4 ready"
+fi
 
 # ntp is installed and running
-
-systemctl is-active ntpd >& /dev/null || systemctl is-active systemd-timesyncd >& /dev/null
-status "ntp or systemd-timesyncd must be installed and active"
+if having_category all api kube node ; then
+    systemctl is-active ntpd >& /dev/null || systemctl is-active systemd-timesyncd >& /dev/null
+    status "ntp or systemd-timesyncd must be installed and active"
+fi
 
 # "persistent" storage class exists in K8s
-
-kubectl get storageclasses |& grep -wq "persistent"
-status "'persistent' storage class should exist in K8s"
+if having_category all kube ; then
+    kubectl get storageclasses |& grep -wq "persistent"
+    status "'persistent' storage class should exist in K8s"
+fi
 
 # privileged pods are enabled in K8s
+if having_category all api ; then
+    kube_apiserver=$(systemctl status kube-apiserver -l | grep "/usr/bin/hyperkube apiserver" )
+    [[ $kube_apiserver == *"--allow-privileged"* ]]
+    status "Privileged must be enabled in 'kube-apiserver'"
+fi
 
-kube_apiserver=$(systemctl status kube-apiserver -l | grep "/usr/bin/hyperkube apiserver" )
-[[ $kube_apiserver == *"--allow-privileged"* ]]
-status "Privileged must be enabled in 'kube-apiserver'"
-
-kubelet=$(systemctl status kubelet -l | grep "/usr/bin/hyperkube kubelet" )
-[[ $kubelet == *"--allow-privileged"* ]]
-status "Privileged must be enabled in 'kubelet'"
+if having_category all node ; then
+    kubelet=$(systemctl status kubelet -l | grep "/usr/bin/hyperkube kubelet" )
+    [[ $kubelet == *"--allow-privileged"* ]]
+    status "Privileged must be enabled in 'kubelet'"
+fi
 
 # dns check for the current hostname resolution
-
-IP=$(host -tA "${SCF_DOMAIN}" | awk '{ print $NF }')
-/sbin/ifconfig | grep -wq "inet addr:$IP"
-status "dns check"
-
+if having_category all api ; then
+    IP=$(host -tA "${SCF_DOMAIN}" | awk '{ print $NF }')
+    /sbin/ifconfig | grep -wq "inet addr:$IP"
+    status "dns check"
+fi
 
 # override tasks infinity in systemd configuration
-
-systemctl cat containerd | grep -wq "TasksMax=infinity"
-status "TasksMax must be set to infinity"
+if having_category all node ; then
+    systemctl cat containerd | grep -wq "TasksMax=infinity"
+    status "TasksMax must be set to infinity"
+fi
 
 exit $FAILED
--- a/cf-deploy-teardown/tasks/k8s-ready-state-check.sh
+++ b/cf-deploy-teardown/tasks/k8s-ready-state-check.sh
@@ -4,7 +4,7 @@ function usage() {
     cat <<EOF
 Usage: $(basename "${0}") [options] [category]
 
-  -h: Displays this help message
+  -h, --help: Displays this help message
 
   Supported categories: all, api, kube, node
   Defaults to: all
@@ -23,6 +23,14 @@ done
 shift $((OPTIND-1))
 
 category="${1:-all}"
+case ${category} in
+    all|api|kube|node)
+	: # ok, nothing to do
+	;;
+    *) usage
+	exit 1
+	;;
+esac
 
 #Script to determine is the K8s host is "ready" for cf deployment
 FAILED=0
--- a/cf-deploy-teardown/tasks/k8s-ready-state-check.sh
+++ b/cf-deploy-teardown/tasks/k8s-ready-state-check.sh
@@ -2,24 +2,33 @@
 
 #Script to determine is the K8s host is "ready" for cf deployment
 FAILED=0
-
 SCF_DOMAIN=${SCF_DOMAIN:-cf-dev.io}
+
 function green() {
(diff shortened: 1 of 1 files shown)


cf-deploy-teardown/tasks/k8s-ready-state-check.sh
```diff
@@ -1,72 +1,126 @@
 #!/bin/bash
 
+function usage() {
+    cat <<EOF
+Usage: $(basename "${0}") ?options? ?category?
```
mook-as: 😀  I'm not that familiar with tcl documentation styles, but I guess this is optional parameters?

cf-deploy-teardown/tasks/k8s-ready-state-check.sh
```diff
@@ -1,72 +1,126 @@
...
+
+  -h: Displays this help message
+
+  Supported categories: all, api, kube, node
+  Defaults to: all
+EOF
+}
+
+while getopts "h" opt; do
```
mook-as: Since you're using getopts anyway, support the long `--help` option too?

cf-deploy-teardown/tasks/k8s-ready-state-check.sh
```diff
@@ -1,72 +1,126 @@
...
+	h)
+	    usage
+	    exit
+	    ;;
+    esac
+done
+
+shift $((OPTIND-1))
+
+category="${1:-all}"
```
mook-as: Do we need better verification here? Say, `k8s-ready-state-check.sh pikachu`?

cf-deploy-teardown/tasks/k8s-ready-state-check.sh
```diff
@@ -1,72 +1,126 @@
...
-status "swapaccount enable"
+    grep -wq "swapaccount=1" /proc/cmdline
+    status "swapaccount enable"
+fi
 
 # docker info should show overlay2
-
-docker info 2> /dev/null | grep -wq "Storage Driver: overlay2"
-status "docker info should show overlay2"
+if having_category all api kube node ; then
```
mook-as: The `kube` environment doesn't actually need a docker, right?
I'm not sure if we care if the api node runs overlay2; that should only be needed for diego-cell anyway, if I understand things correctly? (Which of course runs on node)

cf-deploy-teardown/tasks/k8s-ready-state-check.sh
```diff
@@ -1,72 +1,126 @@
...
-  fi
+    if [ $? -eq 0 ]; then
+	verified "$1"
+    else
+	trouble "$1"
+	FAILED=1
+    fi
 }
+
+function having_category() {
```
mook-as: Optionally, do a `set -- all "$@"` so we don't have to say `all` every time this gets called. Not sure if it's clearer though; just letting you know that's an option.

cf-deploy-teardown/tasks/k8s-ready-state-check.sh
```diff
@@ -1,72 +1,126 @@
...
+    case "$@" in
+	*${category}*)
```
mook-as: Isn't the match here backwards? If category is `kube api`, `having_category api` would test `api` against `*kube api*` which wouldn't match?

cf-deploy-teardown/tasks/k8s-ready-state-check.sh
```diff
@@ -1,72 +1,126 @@
...
+    kubectl get pods --namespace=kube-system --selector k8s-app=kube-dns | grep -Eq '([0-9])/\1 *Running'
+    status "kube-dns should shows 4/4 ready"
+fi
 
 # ntp is installed and running
-
-systemctl is-active ntpd >& /dev/null || systemctl is-active systemd-timesyncd >& /dev/null
-status "ntp or systemd-timesyncd must be installed and active"
+if having_category all api kube node ; then
+    systemctl is-active ntpd >& /dev/null || systemctl is-active systemd-timesyncd >& /dev/null
```
mook-as: If we're running this on _arbitrary_ kube clusters, they may not be using systemd (so `systemctl is-active` would be incorrect).

cf-deploy-teardown/tasks/k8s-ready-state-check.sh
```diff
@@ -1,72 +1,126 @@
...
+    status "ntp or systemd-timesyncd must be installed and active"
+fi
 
 # "persistent" storage class exists in K8s
```
mook-as: Comment needs to be updated.

cf-deploy-teardown/tasks/k8s-ready-state-check.sh
```diff
@@ -1,72 +1,126 @@
...
-
-kube_apiserver=$(systemctl status kube-apiserver -l | grep "/usr/bin/hyperkube apiserver" )
-[[ $kube_apiserver == *"--allow-privileged"* ]]
-status "Privileged must be enabled in 'kube-apiserver'"
-
-kubelet=$(systemctl status kubelet -l | grep "/usr/bin/hyperkube kubelet" )
-[[ $kubelet == *"--allow-privileged"* ]]
-status "Privileged must be enabled in 'kubelet'"
+if having_category all api ; then
+    kube_apiserver=$(systemctl status kube-apiserver -l | grep "/usr/bin/hyperkube apiserver" )
```
mook-as: Arbitrary kubernetes clusters might not be running the API server as the systemd service `kube-apiserver`.  And they might not put hyperkube at `/usr/bin/`.

cf-deploy-teardown/tasks/k8s-ready-state-check.sh
```diff
@@ -1,72 +1,126 @@
...
 
 # dns check for the current hostname resolution
-
-IP=$(host -tA "${SCF_DOMAIN}" | awk '{ print $NF }')
-/sbin/ifconfig | grep -wq "inet addr:$IP"
-status "dns check"
-
+if having_category all api ; then
+    IP=$(host -tA "${SCF_DOMAIN}" | awk '{ print $NF }')
+    /sbin/ifconfig | grep -wq "inet addr:$IP"
```
mook-as: This doesn't make much sense, because
- there might be a load balancer in front of the server (in which case the IP address wouldn't match)
- it's `api.${SCF_DOMAIN}` anyway, right?

cf-deploy-teardown/tasks/k8s-ready-state-check.sh
```diff
@@ -1,72 +1,126 @@
...
+    systemctl is-active ntpd >& /dev/null || systemctl is-active systemd-timesyncd >& /dev/null
+    status "ntp or systemd-timesyncd must be installed and active"
+fi
 
 # "persistent" storage class exists in K8s
-
-kubectl get storageclasses |& grep -wq "persistent"
-status "'persistent' storage class should exist in K8s"
+if having_category all kube ; then
+    test $(kubectl get storageclasses |& wc -l) -gt 1
```
mook-as: Alternatively, `test -n "$(kubectl get storageclasses --output=name 2>/dev/null | grep storageclass/)"` (but that might be too complicated for no gain).

cf-deploy-teardown/tasks/k8s-ready-state-check.sh
andreas-kupries: Yes. Better use `[...]` ?

cf-deploy-teardown/tasks/k8s-ready-state-check.sh
andreas-kupries: How?

cf-deploy-teardown/tasks/k8s-ready-state-check.sh
andreas-kupries: At the moment a bad category will simply not check anything.

cf-deploy-teardown/tasks/k8s-ready-state-check.sh
andreas-kupries: Was not aware of this, will do.

cf-deploy-teardown/tasks/k8s-ready-state-check.sh
andreas-kupries: You specify only a single category to the script, i.e `kube api` is not proper user input.
That is definitely why this way of checking things works and let me avoid using loops for checking.

cf-deploy-teardown/tasks/k8s-ready-state-check.sh
andreas-kupries: See the trello, and what Aaron said. I took his assignments of tests vs categories.

cf-deploy-teardown/tasks/k8s-ready-state-check.sh
mook-as: That's more common in GNU/BSD/Windows style man pages, yes.

cf-deploy-teardown/tasks/k8s-ready-state-check.sh
andreas-kupries: The checks themselves are inherited. Changing them is out of scope for this ticket.
(Aaron's request for the storage class not withstanding).

cf-deploy-teardown/tasks/k8s-ready-state-check.sh
andreas-kupries: Will do.

cf-deploy-teardown/tasks/k8s-ready-state-check.sh
mook-as: Oh! My mistake, I was thinking `getopt`, not `getopts` (yes those are actually different commands 😑 ). Nevermind then.

cf-deploy-teardown/tasks/k8s-ready-state-check.sh
andreas-kupries: See note above.

cf-deploy-teardown/tasks/k8s-ready-state-check.sh
andreas-kupries: This should really discussed with the author of the checks (Prabal?, Hart?)

cf-deploy-teardown/tasks/k8s-ready-state-check.sh
andreas-kupries: I have no idea. See notes above re the checks themselves.

cf-deploy-teardown/tasks/k8s-ready-state-check.sh
mook-as: Right, in that case the `*` in the match isn't necessary.  If the user runs `k8s-ready-state-check.sh noapiplease` it shouldn't match, right?

cf-deploy-teardown/tasks/k8s-ready-state-check.sh
mook-as: Okay, will discuss there

cf-deploy-teardown/tasks/k8s-ready-state-check.sh
andreas-kupries: Done with commit 192fb085f78905f7061fc7c4d7fa586cd40eeca6

cf-deploy-teardown/tasks/k8s-ready-state-check.sh
andreas-kupries: ok

cf-deploy-teardown/tasks/k8s-ready-state-check.sh
andreas-kupries: Done with commit 192fb085f78905f7061fc7c4d7fa586cd40eeca6

cf-deploy-teardown/tasks/k8s-ready-state-check.sh
andreas-kupries: will add validation

andreas-kupries: Done with commit 964f1d71725d2b749e97e998afca91c264b4691c

cf-deploy-teardown/tasks/k8s-ready-state-check.sh
andreas-kupries: Extended help to show that `--help` is indeed supported.
See commit 192fb085f78905f7061fc7c4d7fa586cd40eeca6

cf-deploy-teardown/tasks/k8s-ready-state-check.sh
andreas-kupries: Note, with the new validation added by commit 964f1d71725d2b749e97e998afca91c264b4691c a `noapiplease` will never even reach `having_category`.

cf-deploy-teardown/tasks/k8s-ready-state-check.sh
andreas-kupries: Done with commit 192fb085f78905f7061fc7c4d7fa586cd40eeca6

cf-deploy-teardown/tasks/k8s-ready-state-check.sh
andreas-kupries: Nothing done, keeping current code.

cf-deploy-teardown/tasks/k8s-ready-state-check.sh
andreas-kupries: Done with commit 7dccd13a44f9e5fb20cbd7f7ae39aef4e2366b46

cf-deploy-teardown/tasks/k8s-ready-state-check.sh
andreas-kupries: :( Removed again with commit 1d1732bac035bb8c064d3e001c857aff8e43b7c7

cf-deploy-teardown/tasks/k8s-ready-state-check.sh
prabalsharma: I use "export SCF_DOMAIN=citestxx.name.van"

cf-deploy-teardown/tasks/k8s-ready-state-check.sh
prabalsharma: The scope of this script has changed now. The original scope was to to check our k8s-hosts created by Hart's script and also prove as a requirement for CAASP kube, rather than checking arbitrary kube. Since now this is the case, we need a ticket to "Productize this script" and address these suggestions which Mark pointed out.

PR #15, opened 2017-07-14
Add certstrap pipeline

mook-as:
This adds a concourse pipeline for certstrap, since no upstream binaries.

Since upstream looks completely dormant, I've chosen to use the master branch rather than just tags.


certstrap/tasks/build.sh
```diff
@@ -0,0 +1,14 @@
+#!/bin/sh
+
+set -o errexit -o nounset -o xtrace
+
+export GOPATH=$PWD/go
+
+package="github.com/square/certstrap"
+go build -ldflags='-s -w' "${package}"
```
aarondl: Is this really necessary? How much do we save with this? What if it explodes one day?

certstrap/config-production.yaml
```diff
@@ -0,0 +1,14 @@
+---
+src-repo: https://github.com/square/certstrap
+src-branch: master
+src-private-key: ~
```
aarondl: Is a private key necessary here since this is a public repo and it's already using the https git url?

certstrap/deploy
```diff
@@ -0,0 +1,62 @@
...
+
+if ! test -r "${vars_file}" ; then
+    usage >&2
+    if test -n "${2:-}" ; then
+        printf "Variables file %s is not readable\n" "${vars_file}" >&2
+    fi
+    exit 1
+fi
+
+secrets_file="${PWD}/../../cloudfoundry/secure/concourse-secrets.yml.gpg"
```
aarondl: Still hate this. Surely there's a better way via ENV or something that we use.

certstrap/tasks/build.sh
mook-as: About a meg (something like 2.9M with the args, 4.3 without).  Explosions seem unlikely (it's ldflags), but I can take them out if you really don't like them.

certstrap/config-production.yaml
mook-as: Nope, that's why it's `~`. It's only relevant if it ends up being a private one at some point (i.e. a fork).

certstrap/deploy
mook-as: I guess? I'll think of something.

mook-as: Fixed in 23e9e8f90105f2d3d0dd0c8ad0229acaa68157ab (assumes you can do an envrc or something)

certstrap/tasks/build.sh
mook-as: Actually, `-s` didn't actually save any bytes at all, so took that out in 23e9e8f90105f2d3d0dd0c8ad0229acaa68157ab. `-w` is still there.

PR #22, opened 2017-07-24
Package the helm distribution as well.

andreas-kupries:
Trivial extension. Currently setting up a local concourse for test.

### DIFF
--- a/cf/tasks/package-bits.sh
+++ b/cf/tasks/package-bits.sh
@@ -27,12 +27,12 @@ do
     echo Packaging for $OS, taking $VERSION ...
 
     # Assembling the pieces ...
-    mkdir tmp
+    mkdir -p tmp/kube tmp/helm
 
     # kube configs
-    unzip s3.kube-dist/scf-kube-*.zip -d tmp
+    unzip s3.kube-dist/scf-kube-*.zip -d tmp/kube
     # helm charts
-    unzip s3.helm-dist/scf-helm-*.zip -d tmp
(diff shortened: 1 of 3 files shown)


cf/tasks/package-bits.sh
```diff
@@ -30,8 +30,9 @@ do
     mkdir tmp
 
     # kube configs
-    # helm charts
     unzip s3.kube-dist/scf-kube-*.zip -d tmp
+    # helm charts
+    unzip s3.helm-dist/scf-helm-*.zip -d tmp
```
mook-as: We may need to extract them to `tmp/kube` and `tmp/helm`

andreas-kupries: Hm. IIRC helm has a `helm` sub dir, and the other two `cf` and `uaa`. Should have no conflict.
Regardless, I will check what the concourse gives me.

andreas-kupries: Seems that you are correct. The helm stuff is under `cf` as well, so it mixes with the kube cf.
Now fixing.

andreas-kupries: Done with commit fa59247b7de0d90585ed351095e43e3fe59cd34a

andreas-kupries: Please run this on your concourse and send me the resulting zip archive.
(On my local one it fails to upload to your minio, access denied)
(With a zip file in hand I should be able to complete the docs)

PR #24, opened 2017-07-25
cf: use a custom pool resource that doesn't fail on reset

mook-as:
We were getting useless errors of the form:

```
acquiring lock on: scf-deploy
.......error acquiring lock: exit status 128
```

This was likely the `git fetch` failing; testing the theory.


mook-as:
FWIW, this is using the pool resource with a [single change](https://github.com/concourse/pool-resource/compare/master...mook-as:git-reset-retry).

HartS:
This is no longer relevant because we're not using concourse for scf tests.

PR #26, opened 2017-07-27
Switch from local readiness check to script found in SCF repo.

andreas-kupries:
- Removed local script
- Added resource for scf repo
- Updated task spec and script to use the script from the scf repo/resource.
- Added helper for deploying the pipeline


setup-cf-deploy-test-teardown.sh
```diff
@@ -0,0 +1,24 @@
...
+#           pipelines
+
+if [ ! -f "${secrets}" ]
+then
+    echo -e 1>&2 "$0: Failed to find the secrets file ${secrets}\n\tPlease specify the correct directory holding \"${secfile}\"."
+    exit 1
+fi
+
+fly -t "$target" set-pipeline \
+    -p ${PP}cf-deploy-test-teardown \
```
alexjh: I really like this option for adding a prefix but I think it should have a more descriptive name.

cf-deploy-test-teardown/cf-deploy-test-teardown.yml
```diff
@@ -1,6 +1,14 @@
 # Concourse pipeline to build deploy and teardown cf against a kube cluster
 
 resources:
+- name: src
+  type: git
+  source:
+    uri: https://github.com/SUSE/scf.git
+    branch: develop
+    paths:
+    - bin/*
```
alexjh: The only file it needs to check is `bin/dev/kube-ready-state-check.sh`, I don't see any reason to use a wildcard here.

andreas-kupries: Will fix.

setup-cf-deploy-test-teardown.sh
andreas-kupries: I know I am bad with names. I currently have nothing better. Do you ?

cf-deploy-test-teardown/cf-deploy-test-teardown.yml
andreas-kupries: Done with commit 20b2ceb1881ee5d4882e368998cbf5e2059a40ca

setup-cf-deploy-test-teardown.sh
alexjh: It is one of the two hard problems in computing. :)

How about PIPELINE_PREFIX? I know it's long to type, but I think I'd typically export it once for a testing session, so it's not like you'd be typing it often.

andreas-kupries: True. :+1: on the name, fix coming up ...

andreas-kupries: Done with commit 12fe81f009d764caa3b3846b3bdc62efceee049c

PR #30, opened 2017-08-03
Update helm version

alexjh:


opensuse-kubectl-docker/Dockerfile
```diff
@@ -19,8 +17,11 @@ RUN true \
     curl \
     wget \
     tar \
-    helm \
     direnv \
+    aaa_base bash-completion docker-bash-completion systemd-bash-completion \
```
mook-as: Having to manually install `aaa_base` seems unusual. Why do we need that?

opensuse-kubectl-docker/Dockerfile
```diff
@@ -19,8 +17,11 @@ RUN true \
...
+    && wget -q https://kubernetes-helm.storage.googleapis.com/helm-v2.4.2-linux-amd64.tar.gz -O - | tar xz -C /usr/local/bin --strip-components=1 linux-amd64/helm \
```
mook-as: Since we're getting whatever is the latest `kubectl` anyway… might want to look at https://github.com/kubernetes/helm/blob/e61ee5c54e1d24e791f516cb61059fc910befd1b/scripts/get#L79

Prehasp something liek this would work:
```bash
helm_version=$(curl -sSI https://github.com/kubernetes/helm/releases/latest | awk -F/ '/^Location:/ { print $NF }')
wget -q "https://kubernetes-helm.storage.googleapis.com/helm-${helm_version}-linux-amd64.tar.gz" -O - …
```

opensuse-kubectl-docker/Dockerfile
alexjh: bash completion is set up in `/etc/bash.bashrc`, which is in the aaa_base package.

We can instead:

1. `source /etc/profile.d/bash_completion.sh` in ~/.bashrc
2. Get rid of bash completion altogether. Since this container is meant for interactive usage, I'd rather not do this.

opensuse-kubectl-docker/Dockerfile
alexjh: Yes, that worked, with a minor tweak for CRLF cleanup. There may be a better way to do this in awk, but I think `RS` is the right way.

opensuse-kubectl-docker/Dockerfile
mook-as: Ah, I had assume that `aaa_base` would already been installed (based on its name), but that appears wrong.

PR #31, opened 2017-08-04
rename opensuse cf-ci docker image

prabalsharma:


prabalsharma:
I dont know what's the purpose of cf-ci. The best I can do is change the name of this image from opensuse-cf-ci to something else. Open to suggestions. Aaron was okay with opensuse-cf-ci

aarondl:
The cf-ci image is used to run Concourse tasks relating to building the SCF repo. https://github.com/SUSE/cf-ci/blob/master/cf/tasks/bundle-dist.yml

They could be merged. They could also end up with distinct names if anything since the reality is we've ended up with two images. One that contains things relevant for building and one that contains things relevant to kube/helm orchestration. Since we're already renaming this one let's just change it to cf-ci-orchestration or something of that sort.

PR #34, opened 2017-08-10
Use CONCOURSE_SECRETS_FILE for secrets location

alexjh:


setup-cf-kube-dist.sh
```diff
@@ -1,22 +1,24 @@
...
-# EV 'PIPELINE_PREFIX' = prefix to pipeline name for local customization of test
-#           pipelines
+# EV 'PIPELINE_PREFIX' = prefix to pipeline name for local
+#                        customization of test pipelines
 
-if [ ! -f "${secrets}" ]
-then
-    echo -e 1>&2 "$0: Failed to find the secrets file ${secrets}\n\tPlease specify the correct directory holding \"${secfile}\"."
-    exit 1
+if test -n "${CONCOURSE_SECRETS_FILE:-}"; then
```
mook-as: We should report something sensible when it's not set, rather than attempting to run `gpg -d --no-tty "" 2> /dev/null` and hiding the error message.

setup-cf-kube-dist.sh
```diff
@@ -12,13 +22,16 @@ if test -n "${CONCOURSE_SECRETS_FILE:-}"; then
...
+else
+    echo "ERROR: CONCOURSE_SECRETS_FILE location is not set" >&2
+    exit 3
 fi
 
-fly -t "$target" set-pipeline \
-    -p ${PIPELINE_PREFIX}cf-kube-dist \
-    -c cf-kube-dist/cf-kube-dist.yml \
+fly -t "${target}" set-pipeline \
+    -p "${PIPELINE_PREFIX}${PIPELINE_NAME}" \
```
mook-as: `PIPELINE_PREFIX` may be unset (now that you have `nounset`).

setup-cf-kube-dist.sh
```diff
@@ -1,9 +1,19 @@
...
+#           pipelines
+
+PIPELINE_NAME=cf-kube-dist
+
+if test -z "${1:-}"; then
+    printf "Usage:\n%s <target>\n" "${0}" >&2
+    exit 1
+fi
+
+target="${1}"
```
mook-as: May want to try `${1:-}` and do some error messaging if it's unset, rather than whatever confusing default error we get.

setup-cf-kube-dist.sh
alexjh: Fixed in 5e9b2c4

setup-cf-kube-dist.sh
alexjh: I've updated this to follow what the helm deploy script is doing: if it's not set, use `fly`'s default. If you want to override it, set the target.

setup-cf-kube-dist.sh
alexjh: Or whatever that becomes after I've rebased now...

PR #35, opened 2017-08-15
Add new Jenkins config to deploy SCF on a kube

mook-as:
This assumes the slave has been pre-configured to target kubernetes.


alexjh:
LGTM

PR #36, opened 2017-08-16
Another instance of CONCOURSE_SECRETS_FILE

alexjh:


cf/deploy.rb
```diff
@@ -83,6 +81,14 @@
 end
 
 # Load the secrets file, followed by the local config overrides
+if secrets_path.nil?
+    fail "Neither CONCOURSE_SECRETS_FILE env var nor --secrets-dir option defined"
+end
+
+unless File.exist? secrets_path
```
andreas-kupries: Is it more ruby-ish to say
```
fail "Secrets not found at #{secrets_path}" unless File.exist? secrets_path
```

cf/deploy.rb
```diff
@@ -83,6 +81,14 @@
 end
 
 # Load the secrets file, followed by the local config overrides
+if secrets_path.nil?
```
andreas-kupries: Is it more ruby-ish to say
```
fail "Neither CONCOURSE_SECRETS_FILE env var nor --secrets-dir option defined" if secrets_path.nil?
```

cf/deploy.rb
alexjh: Agreed to both, fixed in 124acc0.

PR #39, opened 2017-09-08
cf: Add running tests (smoke, sats, cats).

mook-as:


alexjh:
Overall it looks good, and it works. Looking for some clarification/comments on some specific things.

PR #45, opened 2017-10-12
Fix tcp routing.

andreas-kupries:
https://trello.com/c/X73Xko3s/366-004tcproutingtestsh-is-failing

We have to set and override the `TCP_DOMAIN` as well.
That will become `properties.acceptance_tests_brain.tcp_domain`,
will become `CF_TCP_DOMAIN` in the test-brain.

### DIFF
--- a/helm-deploy-test/tasks/run-test.sh
+++ b/helm-deploy-test/tasks/run-test.sh
@@ -19,7 +19,8 @@ kube_overrides() {
         obj = YAML.load_file('$1')
         obj['spec']['containers'].each do |container|
             container['env'].each do |env|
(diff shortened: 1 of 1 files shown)


prabalsharma:
yes this was a deployment fault. Jenkins and QA pipelines both used a buggy code.
On my local machine, which I used for manual deployment of HA SCF. The tcp-routing test PASSED cause my test script was replacing all cf.dev.io

`sed` -i "s/cf-dev\.io/${DOMAIN}/g" scf-config/kube/cf/bosh-task/acceptance-tests-brain.yaml`

unlike in pipeline's test script.

PR #46, opened 2017-10-12
Fix same problem in jenkins.

prabalsharma:
tcp_routing test fix

same as https://github.com/SUSE/cf-ci/pull/45

### DIFF
--- a/cf/Jenkinsfile
+++ b/cf/Jenkinsfile
@@ -16,7 +16,9 @@ void runTest(String testName) {
                 obj = YAML.load_file('\$1')
                 obj['spec']['containers'].each do |container|
                     container['env'].each do |env|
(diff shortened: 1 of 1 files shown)


cf/Jenkinsfile
```diff
@@ -16,7 +16,9 @@ void runTest(String testName) {
                 obj = YAML.load_file('\$1')
                 obj['spec']['containers'].each do |container|
                     container['env'].each do |env|
-                        value = env['name'] == 'DOMAIN' ? domain : env['value']
+                        value = env['value']
```
mook-as:
```ruby
env['value'] = case env['name']
    when 'DOMAIN' then domain
    when 'TCP_DOMAIN' then "tcp.#{domain}"
    else env['value'].to_s
end
```

PR #47, opened 2017-10-12
Fix tcp routing in the Jenkinsfile as well. 

andreas-kupries:
We have to set and override the `TCP_DOMAIN` too. That will become `properties.acceptance_tests_brain.tcp_domain`, will become `CF_TCP_DOMAIN` in the test-brain.


prabalsharma:
same fix as https://github.com/SUSE/cf-ci/pull/45

PR #49, opened 2017-10-20
Add Stampy to orchestration image dependencies

HartS:
This will be used for performance data collection


HartS:
I  can fix the conflict and rebase.

I have some incomplete work which uses it, but stampy is *very* simple. You just run `stampy $FILE_NAME $SEQUENCE_NAME $MESSAGE`. This writes a timestamped line in the file passed as the first argument which states that `$MESSAGE` happened in `$SEQUENCE` (sequence is just the thing it pertains to, like `smoke_test`, `deploy` etc.)

The more complex thing is handing off this file from one task to the next, but I found a good way to do that which is in this work also.

PR #51, opened 2017-11-14
CF-USB pipeline

mook-as:
This is the start of the CF-USB pipeline.

Unfortunately, since the CF-USB build system makes quite a few assumptions, this pokes into it a bit.

### DIFF
--- a/cf-usb/cf-usb.yaml
+++ b/cf-usb/cf-usb.yaml
@@ -30,6 +30,7 @@ resources:
   source:
     repository: ((docker-server))((docker-org))/cf-usb-sidecar
     tag: ((docker-tag))
+    tag_as_latest: true
     username: ((docker-username))
     password: ((docker-password))
 - name: docker.buildbase
@@ -37,6 +38,7 @@ resources:
   source:
     repository: ((docker-server))((docker-org))/cf-usb-sidecar-buildbase
     tag: ((docker-tag))
+    tag_as_latest: true
     username: ((docker-username))
     password: ((docker-password))
 - name: docker.mysql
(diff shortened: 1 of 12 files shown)


cf-usb/config-production.yaml
```diff
@@ -0,0 +1,12 @@
+src-repo: https://github.com/SUSE/cf-usb-sidecar.git
+src-branch: develop
+
+ci-repo: https://github.com/SUSE/cf-ci.git
+ci-branch: master
+
+docker-server: ""
+docker-tag: latest
+
+s3-bucket: cf-usb-release-blobstore
```
mook-as: I'm not sure this is the correct bucket to put things into.
Jan thinks we should stick it in our general catch-all bucket until we have something more organized.

cf-usb/cf-usb.yaml
```diff
@@ -30,27 +30,31 @@ resources:
...
     tag: ((docker-tag))
+    tag_as_latest: true
     username: ((docker-username))
     password: ((docker-password))
 - name: docker.mysql-setup
   type: docker-image
   source:
     repository: ((docker-server))((docker-org))/cf-usb-sidecar-dev-mysql-setup
     tag: ((docker-tag))
+    tag_as_latest: true
```
mook-as: This should go on the output params, not resource definition, I think

andreas-kupries: Done with commit ef4a90b749f65afd3166761227f8b96c066f4122. In testing.

andreas-kupries: Works.

PR #60, opened 2017-12-12
Improve domain determination

HartS:
Improve the way the domain for SCF is determined when deploying SCF. This PR changes the way the node IP to use is selected (the previous jsonpath would occasionally select the master IP) by filtering out nodes which have the unschedulable property set.

When running tests, the test script also now retrieves the SCF DOMAIN value from the api pod env, which is more robust.


prabalsharma:
LGTM

prabalsharma:
No jq in continer, had to revert as rc1 is out and I have to test it now.

PR #62, opened 2017-12-18
Copy ceph secret token into new namespace if persistent type is rbd

HartS:


helm-deploy-test/tasks/cf-deploy.sh
```diff
@@ -73,8 +73,12 @@ wait_for_namespace() {
     fi 
 }
 
+PROVISIONER=$(kubectl get storageclasses persistent -o "jsonpath={.provisioner}")
```
prabalsharma: cluster may not use storage class named persistent

prabalsharma:
LGTM

PR #74, opened 2017-12-28
Explicitly name helm releases and improve teardown

HartS:
Explicity naming these will facilitate upgrade testing, as we don't
have to determine the name of the correct release to upgrade in that
task.

### DIFF
--- a/helm-deploy-test/tasks/cf-teardown.sh
+++ b/helm-deploy-test/tasks/cf-teardown.sh
@@ -11,11 +11,9 @@ UAA_NAMESPACE=uaa
 set +o allexport
 
 for namespace in "$CF_NAMESPACE" "$UAA_NAMESPACE" ; do
-    for name in $(helm list --deployed --short --namespace "${namespace}") ; do
-        helm delete "${name}" || true
-    done
-    while kubectl get namespace "${namespace}" ; do
-        kubectl delete namespace "${namespace}" || true
+    kubectl delete namespace "${namespace}" || true
+    while [[ -n $(helm list --short --all ${namespace}) ]]; do
         sleep 10
+        helm delete --purge ${namespace};
     done
 done
--- a/helm-deploy-test/tasks/cf-teardown.sh
+++ b/helm-deploy-test/tasks/cf-teardown.sh
@@ -11,9 +11,9 @@ UAA_NAMESPACE=uaa
 set +o allexport
 
 for namespace in "$CF_NAMESPACE" "$UAA_NAMESPACE" ; do
-    kubectl delete namespace "${namespace}" || true
+    kubectl delete namespace "${namespace}"
     while [[ -n $(helm list --short --all ${namespace}) ]]; do
         sleep 10
-        helm delete --purge ${namespace};
+        helm delete --purge ${namespace} ||:
     done
 done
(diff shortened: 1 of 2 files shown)


helm-deploy-test/tasks/cf-teardown.sh
```diff
@@ -11,11 +11,9 @@ UAA_NAMESPACE=uaa
...
 for namespace in "$CF_NAMESPACE" "$UAA_NAMESPACE" ; do
-    for name in $(helm list --deployed --short --namespace "${namespace}") ; do
-        helm delete "${name}" || true
-    done
-    while kubectl get namespace "${namespace}" ; do
-        kubectl delete namespace "${namespace}" || true
+    kubectl delete namespace "${namespace}" || true
+    while [[ -n $(helm list --short --all ${namespace}) ]]; do
         sleep 10
+        helm delete --purge ${namespace};
```
mook-as: Really tiny nit: do we need the semicolon here? Also, quoting (thought that should be safe)

I believe we had the `|| true` before because it would sometimes fail erroneously (because the machine was too slow and connection to the database times out). Is that no longer a concern?

Do we not need to care about `kubectl delete` failing intermittently?

(I don't actually care about the nits, just the other stuff)

HartS: These are great nits.. not sure how/why I added the semicolon, and the `||:` is probably still a good idea

helm-deploy-test/tasks/cf-teardown.sh
```diff
@@ -11,9 +11,9 @@ UAA_NAMESPACE=uaa
 set +o allexport
 
 for namespace in "$CF_NAMESPACE" "$UAA_NAMESPACE" ; do
-    kubectl delete namespace "${namespace}" || true
+    kubectl delete namespace "${namespace}"
```
HartS: I removed the `|| true` here.. since kube deletions are async, if this fails, something may actually be wrong (api server inaccessible, corrupt kube config). This isn't retried, so we shouldn't `|| true` it.

mook-as: The question then is, _why_ isn't this retried? Sometimes kube API access falls over temporarily (because the machine is being bogged down by the SCF trying to start).

prabalsharma: if we don't use `|| true` then the scrip just exit and pipeline takes it as failure. Which is not true.

HartS: @mook-as this script only runs after passing acceptance tests, so SCF shouldn't be trying to start at that point. I don't see why we'd specifically want a retry here over all the other `kubectl` invocations in [cf-deploy.sh](https://github.com/SUSE/cf-ci/blob/53b8ea561d5f4b1ba492cef10c761ef2df7ca055/helm-deploy-test/tasks/cf-deploy.sh)

mook-as: SCF running could cause things to be slow too, if we had a bug there (but then it'd probably fail CATS).

I _guess_ that's okay…

HartS: I don't think the ability of apiserver to respond should be affected by existing workloads (scf in this case), since the kubelet service isn't running on the master.

PR #75, opened 2018-01-02
Add jobs for HA testing

HartS:


helm-deploy-test/helm-deploy-test.yml
```diff
@@ -36,7 +36,7 @@ resources:
     pool: ((kube-pool-pool))
 
 jobs:
-- name: helm-deploy-test-openSUSE
+- name: helm-deploy-sa-test-openSUSE
```
prabalsharma: did u mean "ha", rather than "sa-"

helm-deploy-test/helm-deploy-test.yml
```diff
@@ -82,6 +82,103 @@ jobs:
   - put: pool.kube-hosts
     params: {release: pool.kube-hosts}
 
+- name: helm-deploy-sa-test-SLES
```
prabalsharma: -ha-

helm-deploy-test/helm-deploy-test.yml
prabalsharma: I guess we should either remove non-ha opensuse or ha-opensuse jobs, we should not use 2 resources(caasp) to test opensue. 1 out of two should be enough. I guess we should keep HA and remove nonHA opensuse job

HartS: sa if for single/standard availability

PR #77, opened 2018-01-05
Increase timeout on helm install to 600

HartS:
The default timeout here is 300; sometimes the release installation
takes longer and we end up hitting the following error:

```
E0105 02:32:08.574555     480 portforward.go:178] lost connection to pod
Error: transport is closing
```


prabalsharma:
@HartS Conflicts

PR #81, opened 2018-01-23
Fix namespace wait check to work with HA

HartS:


helm-deploy-test/tasks/cf-deploy.sh
```diff
@@ -47,8 +47,26 @@ fi
 # Wait until CF namespaces are ready
 is_namespace_ready() {
     local namespace="$1"
-    [[ true == $(kubectl get pods --namespace=${namespace} --output=custom-columns=':.status.containerStatuses[].ready' \
-        | sed '/^$/d' \
+
+    # Create regular expression to match active_passive_pods
```
mook-as: I think this section might be clearer without regex and more shell (untested):
```bash
local -A active_passive_pods=( [diego-api]=0 [diego-brain]=0 [routing-api]=0 )
local all_pod_status="$(kubectl get pods \
    --namespace=${namespace} \
    --output=custom-columns=':.status.containerStatuses[].name,:.status.containerStatuses[].ready')"
local pod_name pod_status
while read -r pod_name pod_status ; do
    if test -n "${active_passive_pods[${pod_name}]}" ; then
        # This is an active passive pod
        active_passive_pods[${pod_name}]="$(( active_passive_pods[${pod_name}] + 1 ))"
    else
        # Not an active/passive pod
        if test "${pod_status}" != "true" ; then
            return 1 # Not ready
        fi
    fi
done <<<"${all_pod_status}"
for pod_status in "${active_passive_pods[@]}" ; do
    if test "${pod_status}" -ne "1" ; then
        return 1 # An active/passive pod does _not_ have exactly one ready instance
    fi
done
```

PR #84, opened 2018-01-26
CI updates for the  sidecar'd mysql broker

andreas-kupries:
https://trello.com/c/MlwUS1BE/537-3-ci-for-mysql-broker

- Redirect generated helm charts to the proper and official location in SUSE s3
- Redirect generated docker images to proper and official SUSE staging registry

General fixups to handle the upheaval brought by the switch from `Godeps` to `vendor`.
Some proactive name changes to make future addition of more brokers easier.

### DIFF
--- a/cf-usb/cf-usb.yaml
+++ b/cf-usb/cf-usb.yaml
@@ -1,4 +1,12 @@
 ---
+resource_types:
+- name: docker-image-resource
+  type: docker-image
+  privileged: true
+  source:
+    repository: concourse/docker-image-resource
+    tag: latest
+
 resources:
 - name: src
   type: git
@@ -26,28 +34,28 @@ resources:
     endpoint: ((s3-endpoint))
     disable_ssl: ((s3-disable-ssl))
 - name: docker.base
-  type: docker-image
+  type: docker-image-resource
   source:
     repository: ((docker-server))((docker-org))/cf-usb-sidecar
     tag: ((docker-tag))
     username: ((docker-username))
     password: ((docker-password))
 - name: docker.buildbase
-  type: docker-image
+  type: docker-image-resource
   source:
     repository: ((docker-server))((docker-org))/cf-usb-sidecar-buildbase
     tag: ((docker-tag))
     username: ((docker-username))
     password: ((docker-password))
 - name: docker.mysql
-  type: docker-image
+  type: docker-image-resource
   source:
     repository: ((docker-server))((docker-org))/cf-usb-sidecar-mysql
     tag: ((docker-tag))
     username: ((docker-username))
     password: ((docker-password))
 - name: docker.mysql-setup
-  type: docker-image
+  type: docker-image-resource
   source:
     repository: ((docker-server))((docker-org))/cf-usb-sidecar-mysql-setup
     tag: ((docker-tag))
@@ -142,7 +150,7 @@ jobs:
     - put: docker.mysql
       params:
         build: mysql-binary
-        load_base: docker.base-fixed
+        load_bases: [ docker.base-fixed, docker.buildbase-fixed ]
         tag_as_latest: true
         tag: mysql-binary/tag
     - put: docker.mysql-setup
--- a/cf-usb/cf-usb.yaml
+++ b/cf-usb/cf-usb.yaml
@@ -71,23 +71,29 @@ jobs:
       trigger: true
     - get: ci
       trigger: true
-  - aggregate:
-    - do:
-      - task: generate
-        file: ci/cf-usb/tasks/csm-generate.yml
-      - task: build
-        file: ci/cf-usb/tasks/csm-build.yml
-        output_mapping:
-          out: csm-binary
-      - put: docker.base
-        params:
-          tag_as_latest: true
-          build: csm-binary
-    - put: docker.buildbase
-      params:
-        tag_as_latest: true
-        build: src
-        dockerfile: src/scripts/docker/Dockerfile-build
+  - task: generate
+    file: ci/cf-usb/tasks/csm-generate.yml
+  - task: build
+    file: ci/cf-usb/tasks/csm-build.yml
+    output_mapping:
+      out: csm-binary
+  - put: docker.buildbase
+    get_params: {save: true}
+    params:
+      tag_as_latest: true
+      build: src
+      dockerfile: src/scripts/docker/Dockerfile-build
+  - task: fix-buildbase
+    file: ci/cf-usb/tasks/fix-base.yml
+    input_mapping:
+      in: docker.buildbase
+    output_mapping:
+      out: docker.buildbase-fixed
+  - put: docker.base
+    params:
+      load_base: docker.buildbase-fixed
+      tag_as_latest: true
+      build: csm-binary
 - name: sidecar-mysql
   plan:
   - aggregate:
@@ -148,7 +154,7 @@ jobs:
         tag: mysql-binary/tag
     - put: s3.mysql
       params:
-        file: mysql-helm/cf-usb-*.tgz
+        file: mysql-helm/cf-usb-sidecar-mysql*.tgz
   - put: semver.cf-usb-sidecar-mysql
     params:
       pre: pre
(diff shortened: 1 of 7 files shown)


cf-usb/cf-usb.yaml
```diff
@@ -1,4 +1,12 @@
 ---
+resource_types:
+- name: docker-image-resource
```
mook-as: Please add a comment about why we need this (and ideally, when we can remove it and use the built-in version).

Also, I'm pretty sure you can just call this `docker-image` and override the old version and it should still work.

cf-usb/cf-usb.yaml
```diff
@@ -136,7 +150,7 @@ jobs:
     - put: docker.mysql
       params:
         build: mysql-binary
-        load_base: docker.base-fixed
+        load_bases: [ docker.base-fixed, docker.buildbase-fixed ]
```
mook-as: This would probably look better in the other list format:
```yaml
load_bases:
- docker.base-fixed
- docker.buildbase-fixed
```

cf-usb/config-production.yaml
```diff
@@ -4,12 +4,14 @@ src-branch: develop
 ci-repo: https://github.com/SUSE/cf-ci.git
 ci-branch: master
 
-docker-server: ""
 docker-tag: latest
+docker-server: "staging.registry.howdoi.website/"
```
mook-as: That should be in the config file, not here.

cf-usb/tasks/csm-build.sh
```diff
@@ -14,7 +14,7 @@ cd "${SIDECAR_ROOT}"
 if ! test -d vendor ; then
```
mook-as: Would this ever be true again? I believe it was put in as a transition for the move into `vendor/`.

cf-usb/tasks/csm-generate.sh
```diff
@@ -8,8 +8,8 @@ cd "${SIDECAR_ROOT}"
 if ! test -d vendor ; then
```
mook-as: (Same as above)

cf-usb/tasks/mysql-build.sh
```diff
@@ -2,8 +2,9 @@
 set -o nounset -o errexit -o xtrace
 export GOPATH="${PWD}"
 export START_DIR="${PWD}"
+service=mysql
```
mook-as: Perhaps this can be a [run arg](http://concourse.ci/running-tasks.html#task-run-args) or [environment variable](http://concourse.ci/running-tasks.html#task-params) from the pipeline, so we only have one script for all sidecars?

cf-usb/cf-usb.yaml
andreas-kupries: Well, I can try.

cf-usb/cf-usb.yaml
andreas-kupries: I had a syntax error when I tried that at first.
Spaces vs tabs ?

cf-usb/config-production.yaml
andreas-kupries: This is a/the `config` file. Or are you talking `cf-usb.yml`, i.e. the pipeline definition instead ?
If yes, where in that definition ?

cf-usb/cf-usb.yaml
mook-as: I believe tabs are banned in YAML, for Reasons.

cf-usb/config-production.yaml
mook-as: Oh, so it is. Nevermind then, I got confused, sorry. (Unless we want to stick it in the secrets file)

cf-usb/tasks/mysql-build.sh
andreas-kupries: Yes, this chunk was preparation for things like that, without knowing how to specify such.
Thanks to the links I can do that.

cf-usb/tasks/csm-build.sh
andreas-kupries: Well, it could happen for a test pipeline pointed to an old commit.

andreas-kupries: But I am fine with removing this entirely.

cf-usb/tasks/csm-generate.sh
andreas-kupries: (See above)

cf-usb/cf-usb.yaml
andreas-kupries: Heh. Will change.

cf-usb/config-production.yaml
andreas-kupries: 👍

cf-usb/tasks/mysql-build.sh
andreas-kupries: Would we still need separate task yml files ?

andreas-kupries: Looks to be. Done.

cf-usb/cf-usb.yaml
andreas-kupries: Done with commit 345074c906484e2478bf6d1289bd535795c298ff, in testing.

cf-usb/cf-usb.yaml
andreas-kupries: Done with commit 345074c906484e2478bf6d1289bd535795c298ff, in testing.

cf-usb/config-production.yaml
andreas-kupries: Done with commit 345074c906484e2478bf6d1289bd535795c298ff, in testing.

cf-usb/tasks/csm-build.sh
andreas-kupries: Done with commit 345074c906484e2478bf6d1289bd535795c298ff, in testing.

cf-usb/tasks/csm-generate.sh
andreas-kupries: Done with commit 345074c906484e2478bf6d1289bd535795c298ff, in testing.

cf-usb/tasks/mysql-build.sh
andreas-kupries: Done with commit 345074c906484e2478bf6d1289bd535795c298ff, in testing.

cf-usb/config-production.yaml
mook-as: So it turns out we never set the correct registry in the (helm) configs, so we can't find it.

cf-usb/cf-usb.yaml
```diff
@@ -136,7 +156,9 @@ jobs:
     - put: docker.mysql
       params:
         build: mysql-binary
-        load_base: docker.base-fixed
+        load_bases:
+        - docker.base-fixed
+        - docker.buildbase-fixed
         tag_as_latest: true
         tag: mysql-binary/tag
```
mook-as: The tag doesn't seem to match what we actually put in the helm config; I can't pull it manually.

cf-usb/config-production.yaml
andreas-kupries: Aaaah, now I understand (on 3rd re-read). The build script has to set the `DOCKER_REPOSITORY` to the one chosen by the config so that the chart has the correct repo baked in. Either that, or subst the correct value into the chart before packaging it into the tarball.

(If we need docker hub during the chart generation)

cf-usb/cf-usb.yaml
andreas-kupries: Should be fixed with commit 293ec24c790b75cc444f7d32a793d593f16d8577

andreas-kupries:
Db image fixed with commit c122d5a4ec56227e53eb75c4757bcd94cef16b24.
Note, this has an associated commit/branch in the cf-usb-sidecar project (New `Dockerfile`).

andreas-kupries:
Additional fix with commit f702f3e4c4be9af6d73f45e0698cdedec619d141
Test pipeline works.
Note that the test pipeline uses branch `MlwUS1BE-ci-mysql-broker` of the sidecar repo to have access to the new `Dockerfile-db`.

PR #86, opened 2018-02-06
Minor cleanups for the cf-usb pipeline

mook-as:
Each commit is a separate thing:

- Don't pull the final images after pushing them; we don't use them for anything
- Merge the postgres and mysql build task configs, since the delta is trivial
- Make the base image building slightly more parallel


cf-usb/tasks/service-build.yml
```diff
@@ -15,6 +15,6 @@ outputs:
 - name: helm-out
 - name: docker-out
 run:
-  args:
-  - mysql
   path: ci/cf-usb/tasks/service-build.sh
+params:
+  SERVICE: ~
```
andreas-kupries: Does it need this ? Would it need `DESTINATION` as well ?

mook-as: Nah, it actually needs neither (because they get overridden from the pipeline). I'm just putting it here as a reference for people looking at reusing the task next time (e.g. for a redis service).

andreas-kupries: I see. Ok.

PR #87, opened 2018-02-08
CF-USB: Build against SLE base

mook-as:
This gives enough configurability to build against a SLE base image.

See also https://github.com/SUSE/cf-usb-sidecar/pull/11

### DIFF
--- a/cf-usb/deploy
+++ b/cf-usb/deploy
@@ -5,7 +5,7 @@ set -o nounset -o errexit
 if test -z "${1:-}" -o ! -r "config-${1:-}.yaml" ; then
     exec >&2
     echo "Failed to find configuration file config-${1:-}.yaml"
-    echo "Usage: ${0} <config-variant>"
+    echo "Usage: ${0} <config-variant> [pipeline-suffix]"
     exit 1
 fi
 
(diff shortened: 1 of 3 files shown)


cf-usb/deploy
```diff
@@ -9,13 +9,13 @@ if test -z "${1:-}" -o ! -r "config-${1:-}.yaml" ; then
...
 
-alt_secrets_file="CONCOURSE_SECRETS_FILE_${1}"
+alt_secrets_file="CONCOURSE_SECRETS_FILE_${1//-/_}"
 if test -n "${!alt_secrets_file:-}" -a -r "${!alt_secrets_file:-}" ; then
     CONCOURSE_SECRETS_FILE="${!alt_secrets_file:-}"
 fi
 
 fly set-pipeline \
-    --pipeline="cf-usb" \
+    --pipeline="cf-usb${2:+-${2:-}}" \
```
andreas-kupries: Nit: This argument is not mentioned in the usage text.

mook-as: Fixed in b67345d05d95054d15025239e8bcacc4b99be90d

PR #88, opened 2018-02-09
Prabal upgrades ci

prabalsharma:
https://ci.howdoi.website/teams/main/pipelines/prabal-cap-qa-upgrades/jobs/cap-qa-upgrade-SA-SLES/builds/4 seems be working. tests fails because of other reasons, unrelated to this ticket.


helm-deploy-test/tasks/cf-deploy.sh
```diff
@@ -20,12 +20,20 @@ UAA_PORT=2793
 
 CF_NAMESPACE=scf
 UAA_NAMESPACE=uaa
+CAP_DIRECTORY=s3.scf-config
 set +o allexport
 
-unzip s3.scf-config/scf-*.zip -d s3.scf-config/
+# For upgrades tests
+if [ -n "${CAP_INSTALL_VERSION:-}" ]; then
```
HartS: `:-` isn't necessary in this expansion

prabalsharma: I just used the format we are already using https://github.com/SUSE/cf-ci/blob/master/helm-deploy-test/tasks/cf-deploy.sh#L34-L43
I guess it should be another PR to address all such changes

HartS: pretty sure I made the same comment back then. You should avoid putting things into code that serve no function, unless they're making the code more readable. Consider the following two snippets:

```
if [ -n ${SOME_FLAG} ]; then
  echo "SOMEFLAG was set"
fi
```

vs.


```
if [ -n ${SOME_FLAG} ]; then
  if true;
    echo "SOMEFLAG was set"
  fi
fi
```

Since the two snippets are functionally equivalent, but the latter one contains unnecessary characters which make it less straightforward, you should prefer the first form to the latter.


In this case, testing that ${VAR:-} is functionally equivalent to testing that ${VAR} is set, but this is less apparent then the example above (due to the different types of variable expansions), so it causes even more unnecessary confusion.

HartS:
I added several changes to this branch, basically to avoid reaching into tasks from a different pipeline path. Instead of putting config which can be reused between both pipelines in two places, I put it into a `qa-pipelines/` directory and symlink the config names in both pipeline directories (which were also moved into `qa-pipelines/` to the config files. That we we don't need to make changes in multiple places in the future.

The tasks files from both pipelines were moved into `qa-pipelines/tasks/` and references in both pipelines' yml files were changed to correctly reference them.

Also resolved an issue with the vancouver config (even though we're not using it right now) with inconsistent param names.

PR #91, opened 2018-02-21
Work around pv stuck termination issue

HartS:


HartS:
Got an SS deletion stuck when tearing down an HA deployment.. may need further investigation

PR #94, opened 2018-03-08
fix secret in test yaml

prabalsharma:


qa-pipelines/tasks/run-test.sh
```diff
@@ -41,12 +50,16 @@ container_status() {
 }
 
 image=$(awk '$1 == "image:" { gsub(/"/, "", $2); print $2 }' "${CAP_DIRECTORY}/kube/cf${CAP_CHART}/bosh-task/${TEST_NAME}.yaml")
+
+remove_test_config () { rm "${kube_yaml}" ; }
```
HartS: This may be unnecessary because it's running in a concourse container anyway

PR #95, opened 2018-03-13
Harts test app downtime during upgrades

HartS:


qa-pipelines/tasks/cf-upgrade.sh
```diff
@@ -135,3 +176,19 @@ helm upgrade scf ${CAP_DIRECTORY}/helm/cf${CAP_CHART}/ \
...
+# Kill it, so we don't get any messages about the app becoming unreachable at the end
+if jobs %% &>/dev/null; then
+  echo "Terminating app monitoring background job"
+  kill %1
+fi
+
+echo "Results of app monitoring:"
+uniq -c "${monitor_file}"
+cf delete -f go-env
+cf delete-org -f go-env
```
prabalsharma: org is `testorg`

qa-pipelines/tasks/cf-upgrade.sh
```diff
@@ -98,7 +98,48 @@ wait_for_namespace() {
...
+  local app_url=$1
+  local log_file=$2
+  local count=0
+  local last_state=
+  local new_state=
+  echo "monitoring URL ${app_url}"
+  while true; do
+    new_state=$({ curl -sI "${1}" || echo "URL could not be reached"; } | head -n1 | tee -a "${log_file}")
+    if [[ "${new_state}" != "${last_state}" ]]; then
+      echo "state change for ${1}: ${new_state}"
```
prabalsharma: (suggestion) 
```monitoring URL http://go-env.10.84.101.213.omg.howdoi.website
state change for http://go-env.10.84.101.213.omg.howdoi.website: HTTP/1.1 200 OK
Release "uaa" has been upgraded. Happy Helming!```

it will be nice if we `last_state` is also shown
` echo "state change for ${1}: ${new_state} from ${last_state}"`

File: qa-pipelines/tasks/cf-upgrade.sh [hunk 3]
```diff
@@ -135,3 +176,19 @@ helm upgrade scf ${CAP_DIRECTORY}/helm/cf${CAP_CHART}/ \
...
+done
+
+# If we get here because the app is ready, monitor_url will still be running in the background
+# Kill it, so we don't get any messages about the app becoming unreachable at the end
+if jobs %% &>/dev/null; then
+  echo "Terminating app monitoring background job"
+  kill %1
+fi
+
+echo "Results of app monitoring:"
```
prabalsharma: (suggestion) 
i think we can add one more echo line here:
echo "Total seconds : App State"

qa-pipelines/tasks/cf-upgrade.sh
```diff
@@ -98,7 +98,48 @@ wait_for_namespace() {
...
+# If the application state changes, print this to stdout as well
+monitor_url() {
+  local app_url=$1
+  local log_file=$2
+  local count=0
+  local last_state=
+  local new_state=
+  echo "monitoring URL ${app_url}"
+  while true; do
+    new_state=$({ curl -sI "${1}" || echo "URL could not be reached"; } | head -n1 | tee -a "${log_file}")
```
prabalsharma: we can call "Downtime" instead of "URL could not be reached". The summary will become more aligned for this "Downtime" test. (suggestion)

qa-pipelines/tasks/cf-upgrade.sh
HartS: I can't think of an easy way to say this that will make sense with the first state change (when last state is empty), without introducing additional logic around that. Anyway, the last state can be seen by scrolling back to the previous announcement of state change.

qa-pipelines/tasks/cf-upgrade.sh
HartS: There are various kinds of downtime though.. even if the router is up, if the app is still down you'll see a 404.. Since there's not just one kind of downtime, we want this output to inform the user that the app was actually *unreachable* (so there was no http response status), as opposed to various other types of responses that would *also* indicate downtime.

PR #97, opened 2018-03-23
Change tasks to get ca cert from the right secret

HartS:


qa-pipelines/tasks/cf-upgrade.sh
```diff
@@ -148,15 +148,32 @@ helm upgrade uaa ${CAP_DIRECTORY}/helm/uaa${CAP_CHART}/ \
...
 
-# Determine the Helm revision of the chart controlling the specified namespace.
-helm_revision() { helm list --date --reverse --max 1 --namespace "$1" | awk '{ print $2 }' | tail -n 1 ; }
-get_uaa_secret () {
-    kubectl get secret secret-$(helm_revision "${UAA_NAMESPACE}") \
-    --namespace "${UAA_NAMESPACE}" \
-    -o jsonpath="{.data['$1']}"
+
+function semver_is_gte() {
+  # Returns successfully if the left-hand semver is greater than or equal to the right-hand semver
```
prabalsharma: # Returns successfully if the left-hand semver (symmetrical version)  is greater than or equal to the right-hand semver

prabalsharma: Returns successfully if the left-hand semver (symmetrical version) is greater than or equal to the right-hand semver

prabalsharma: Also can u add the logic example you were explaining 10.1.2 vs 2.3.4

HartS: semantic version.. but this is a well-known thing: https://semver.org/

qa-pipelines/tasks/cf-deploy.sh
```diff
@@ -114,15 +114,31 @@ helm install ${CAP_DIRECTORY}/helm/uaa${CAP_CHART}/ \
...
 wait_for_namespace "${UAA_NAMESPACE}"
 
-# Determine the Helm revision of the chart controlling the specified namespace.
-helm_revision() { helm list --date --reverse --max 1 --namespace "$1" | awk '{ print $2 }' | tail -n 1 ; }
-get_uaa_secret () {
-    kubectl get secret secret-$(helm_revision "${UAA_NAMESPACE}") \
-    --namespace "${UAA_NAMESPACE}" \
-    -o jsonpath="{.data['$1']}"
+function semver_is_gte() {
+  # Returns successfully if the left-hand semver is greater than or equal to the right-hand semver
```
prabalsharma: Returns successfully if the left-hand semver (symmetrical version) is greater than or equal to the right-hand semver

prabalsharma: Also can u add the example of 10.1.2. vs 2.4.3 on comment. (Suggestion)

qa-pipelines/tasks/cf-upgrade.sh
```diff
@@ -148,15 +148,18 @@ helm upgrade uaa ${CAP_DIRECTORY}/helm/uaa${CAP_CHART}/ \
...
-helm_revision() { helm list --date --reverse --max 1 --namespace "$1" | awk '{ print $2 }' | tail -n 1 ; }
-get_uaa_secret () {
-    kubectl get secret secret-$(helm_revision "${UAA_NAMESPACE}") \
-    --namespace "${UAA_NAMESPACE}" \
-    -o jsonpath="{.data['$1']}"
+# Get the version of the helm chart for uaa
+helm_chart_version() { grep "^version:"  ${CAP_DIRECTORY}/helm/uaa${CAP_CHART}/Chart.yaml  | sed 's/version: *//g' ; }
+generated_secrets_secret() { kubectl get --namespace "${UAA_NAMESPACE}" secrets --output "custom-columns=:.metadata.name" | grep -F "secrets-$(helm_chart_version)-" | sort | tail -n 1 ; }
+get_internal_ca_cert() {
+    local uaa_secret_name=$(generated_secrets_secret)
```
prabalsharma: I think this will we good to include for any upgrade tests required prior to 273
+    local uaa_secret_name
+    if semver_is_gte ${helm_chart_version} 2.7.3; then
+        uaa_secret_name=$(generated_secrets_secret)
+    else
+        uaa_secret_name=secret
+    fi

prabalsharma: actually we dont care about anything below 270. So this should be okay

qa-pipelines/tasks/cf-upgrade.sh
```diff
@@ -194,5 +194,6 @@ fi
 echo "Results of app monitoring:"
 echo "SECONDS|STATUS"
 uniq -c "${monitor_file}"
+cf login -u admin -p changeme
```
prabalsharma: you will have to select org and space as well

PR #98, opened 2018-03-26
Add cf-usb-plugin CI

alexjh:


cf-usb-plugin/config-production.yaml
```diff
@@ -0,0 +1,21 @@
+# Production configuration for the cf-usb-plugin pipelines
+# Note that the secrets are not here; they are in $CONCOURSE_SECRETS_FILE
+
+# should be upstream: cloudfoundry/cli-plugin-repo
+upstream: SUSE/cli-plugin-repo
```
andreas-kupries: Upstream for the production config is not SUSE.

alexjh: Fixed in 6a50579

cf-usb-plugin/config-production.yaml
```diff
@@ -0,0 +1,21 @@
...
+src-branch: develop
+
+ci-repo: https://github.com/SUSE/cf-ci.git
+ci-branch: master
+
+s3-endpoint: ~
+s3-access-key: *aws-access-key
+s3-secret-key: *aws-secret-key
+s3-bucket: cf-usb-plugin-release
+s3-prefix: /
```
andreas-kupries: Looking at this where exactly is this uploaded to ?

Looking at
https://concourse.ca-west-1.howdoi.website/teams/main/pipelines/ajh-cf-usb-plugin-release/jobs/release/builds/6
the path is shown as `/cf-usb-plugin-0.0.2-release.tgz` and that does not seem to be correct.

Had trouble finding it. Then went back to the backup pipeline. The path there is `cf-plugin-backup/cf-plugin-backup-1.0.5-release.tgz`, which translated to the s3 path `s3://cf-opensusefs2/cf-plugin-backup/cf-plugin-backup-1.0.5-release.tgz`

With this I managed to find the upload from here at `s3://cf-usb-plugin-release/cf-usb-plugin-0.0.2-release.tgz`

This looks to be a new toplevel bucket. Is that the bucket we want?

alexjh: I think it's a reasonable place to put it. I don't think `cf-opensusefs2` is the right place because it's starting to become a catch-all rather than its original use as files for the OpenSUSE stack.

Maybe the bucket name should lose the `-release` suffix and just be `cf-usb-plugin`?

andreas-kupries: I am not sure about namespacing here. With the `cf-opensusefs2` it is clear to be something of SUSE's. With the proposed names it sort of looks like an official CF bucket ?!

Should we pull in Troy or Gary ?

alexjh: Ah, I see where you're coming from. How about `suse-cf-usb-plugin`?

S3 bucket names are a bit of a Wild West situation... I'm surprised we haven't seen people taking bucket names and reselling them yet. :) Although I'm not too concerned about it because these are all internal to our pipelines; the final release will be coming from `github.com/SUSE/cf-usb-plugin/releases/...`

gaktive: Troy likes what Alex suggested: `cf-usb-plugin`
Most users would rely on Github for the release anyways.

PR #101, opened 2018-04-19
Commit sidecar sources to obs for legal review

thardeck:


cf-usb/tasks/service-commit-sources.sh
```diff
@@ -0,0 +1,36 @@
+#!/usr/bin/env bash
+
+# This script publishes the cf-usb-sidecar bundle
+# Most arguments come from the task definition
+
+set -o errexit
+
+if [[ "${COMMIT_SOURCES}" != "true" ]]; then
+  echo "+------------------------------------------------+"
+  echo "| On openSUSE builds no source will be published |"
```
aduffeck: This message can be confusing if the flag is set to `false` for non-openSUSE builds in the future.

cf-usb/tasks/service-commit-sources.sh
```diff
@@ -0,0 +1,36 @@
...
+osc checkout -M "${OSC_BASE_PATH}"
+pushd "${OSC_BASE_PATH}"
+if [ ! -d "${PACKAGE}" ]; then
+  osc mkpac "${PACKAGE}"
+fi
+pushd "${PACKAGE}"
+tar cfz "cf-usb-sidecar.tar.gz" -C "../../../src/" ./
+osc addremove
+osc commit -m "Add sources"
+popd
```
aduffeck: there's no real point for pushing the paths and popping them at the end of the script, no?

PR #102, opened 2018-04-23
Use different pull request resource

thardeck:
to disable the automatic test runs against forks.
https://github.com/jtarchie/github-pullrequest-resource

Only people with write access should be able to trigger automatic tests.


mook-as:
Looks like this change is already deployed on the pipeline. (That's fine, just means I didn't have to re-deploy.)

PR #104, opened 2018-04-26
Use github token for hub

thardeck:
instead of password to prevent creation of tokens each time.


cf-usb/config-production.yaml
```diff
@@ -28,7 +28,7 @@ s3-disable-ssl: false
 charts-release-tool-url: *cap-release-tool-url
 charts-github-org: SUSE
 charts-github-username: *github-username
-charts-github-password: *github-password
+charts-github-access-token: *github-acces-token
```
mook-as: Typo :(

PR #106, opened 2018-05-07
Reworked version processing in the chart builder

andreas-kupries:
Put only the main version information into the chart.
The archive will still use the full version for its filename.

Ref: https://trello.com/c/u18ks0c3/632-2-sidecars-switch-tagging-of-current-builds
For the test pipeline see reference.


andreas-kupries:
I see. Ok. Will change.

andreas-kupries:
Actually that means I have to undo the ci change, and do a sidecar sources change.
That means we can close this PR instead, and open a new one for cf-usb-sidecar.

mook-as:
Closing, fixing via https://github.com/SUSE/cf-usb-sidecar/pull/18 instead

PR #107, opened 2018-05-10
2.9.0 secrets workaround

prabalsharma:
disable brain-tests for 2.8.0, remove trailing `/` from registry hostname, kube.secrets_generation_counter=2


qa-pipelines/tasks/cf-upgrade.sh
```diff
@@ -179,6 +179,7 @@ helm upgrade scf ${CAP_DIRECTORY}/helm/cf${CAP_CHART}/ \
     --set "env.UAA_HOST=${UAA_HOST}" \
     --set "env.UAA_PORT=${UAA_PORT}" \
     --set "secrets.UAA_CA_CERT=${CA_CERT}" \
+    --set "kube.secrets_generation_counter=2"\
```
HartS: We need to do this for  UAA also

PR #109, opened 2018-05-28
Harts cf usb test in upgrade

HartS:


HartS:
For SA upgrades, the usb test apps may not yet be available when it gets to https://github.com/SUSE/cf-ci/blob/958b6228aeb1fc6bd0acc142a0f7b0ff3f71ceed/qa-pipelines/tasks/usb-post-upgrade.sh, so we may need to add some retry logic.

PR #118, opened 2018-06-21
Use current branch for pipeline deploy

HartS:
No longer specify the branch in the config file. The user can specify it
when deploying (with the '-b' or '--branch' flags). Otherwise, the
set-pipeline script will take the name of the current branch. If the
branch doesn't exist on the remote, or the user is in a detached head,
provide appropriate output and exit


HartS:
Added a commit to append the branch/tag name to the pipeline name, which might be used instead of the prefix option.

For example, testing this work by deploying the `cap-qa-deployments` pipeline from this branch results in the pipeline named `cap-qa-deployment-harts-add-branch-param`

PR #124, opened 2018-06-28
Harts get domain from configmap

HartS:
Uses `cap-values` configmap to get information for `external_ips` and `DOMAIN` when deploying CAP. This will be set by the `prep-new-cluster.sh` script in the future (which will be compatible with CaaSP2, CaaSP3, Azure, and Openstack), but for now this can be added manually with a command like the following: `kubectl create configmap -n kube-system cap-values --from-literal=internal-ip=172.... --from-literal=public-ip=10....`

Builds on top of https://github.com/SUSE/cf-ci/pull/123

### DIFF
--- a/qa-pipelines/tasks/cf-deploy-upgrade-common.sh
+++ b/qa-pipelines/tasks/cf-deploy-upgrade-common.sh
@@ -149,7 +149,6 @@ set -o allexport
 # the public IP (used for DOMAIN) will be taken from the floating IP or load balancer IP.
 external_ip=$(kubectl get configmap -n kube-system cap-values -o json | jq -r '.data["internal-ip"]')
 public_ip=$(kubectl get configmap -n kube-system cap-values -o json | jq -r '.data["public-ip"] // empty')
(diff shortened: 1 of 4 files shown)


qa-pipelines/tasks/cf-deploy-upgrade-common.sh
```diff
@@ -0,0 +1,161 @@
...
+}
+
+set -o allexport
+
+# The internal/external and public IP addresses are now taken from the configmap set by prep-new-cluster
+# The external_ip is set to the internal ip of a worker node. When running on openstack or azure, 
+# the public IP (used for DOMAIN) will be taken from the floating IP or load balancer IP.
+external_ip=$(kubectl get configmap -n kube-system cap-values -o json | jq -r '.data["internal-ip"]')
+public_ip=$(kubectl get configmap -n kube-system cap-values -o json | jq -r '.data["public-ip"] // empty')
+public_ip=${public_ip:-$external_ip}
```
prabalsharma: I guess this should be removed

PR #126, opened 2018-06-28
Add new configs for ecp and azure pools

HartS:


qa-pipelines/config-provo.yml
```diff
@@ -28,4 +28,4 @@ organization: splatform
 magic-dns-service: omg.howdoi.website
 
 cap-sle-url: https://s3.amazonaws.com/cap-release-archives/master/scf-sle-2.8.0%2Bcf1.15.0.0.g5aa8b036-amd64.zip
```
prabalsharma: should be 2.10.1

PR #133, opened 2018-07-03
Add pipeline for charts that work with minibroker

alexjh:


minibroker/minibroker-master.yaml
```diff
@@ -0,0 +1,293 @@
...
+    <<: *src-ci-configs
+    paths:
+    - minibroker
+
+- name: src-ci-docker
+  type: git
+  source:
+    <<: *src-ci-configs
+    paths:
+    - minibroker/Dockerfile
```
mook-as: Why don't we need this in the -check pipeline?

minibroker/minibroker-check.yaml
```diff
@@ -0,0 +1,451 @@
...
+      put: status
+      params:
+        context: generate-minibroker-chart
+        description: generate-minibroker-chart
+        path: src
+        state: success
+- name: generate-minibroker-repo
+  plan:
+  - aggregate:
+    - get: src
```
mook-as: Is this used?

minibroker/scripts/build-minibroker.sh
```diff
@@ -0,0 +1,15 @@
...
+
+# Prepare minibroker dir for building docker image
+cp -R src minibroker/
+
+# Build the executable
+mkdir -p gopath/src/github.com/osbkit
+cp -R src gopath/src/github.com/osbkit/minibroker
+export GOPATH=$PWD/gopath
+cd gopath/src/github.com/osbkit/minibroker
+make build-linux
```
mook-as: `make -C gopath/src/github.com/osbkit/minibroker build-linux`?

minibroker/scripts/generate-minibroker-repo.sh
```diff
@@ -0,0 +1,5 @@
+#!/usr/bin/env sh
+set -e
+set -x
+cp s3.*chart/*.tgz repo
+helm repo index --url https://minibroker-helm-charts.s3.amazonaws.com/minibroker-charts repo
```
mook-as: Isn't the s3 bucket name supposed to be customizable?

minibroker/setup-bucket.sh
```diff
@@ -0,0 +1,16 @@
...
+set -x
+
+s3cmd mb s3://minibroker-helm-charts
+
+aws s3api put-bucket-versioning --bucket minibroker-helm-charts --versioning-configuration Status=Enabled
+
+cat <<EOF > index.yaml
+apiVersion: v1
+entries: ~
+generated: 2018-06-28T21:56:27.538677094Z
```
mook-as: Using an obviously-fake initial timestamp here would let people know that the empty file is because it hasn't ran before, not because it ran and gave nothing. (1970 is a nice year)

minibroker/scripts/build-minibroker.sh
alexjh: Fixed in 8dea0af

minibroker/setup-bucket.sh
alexjh: Fixed in 0442755

minibroker/minibroker-check.yaml
alexjh: It's used by the `status` to get the commit for the PR. The Github API is called to set the status for that commit.

minibroker/minibroker-master.yaml
alexjh: The check pipeline is using the docker image generated in the master pipeline. It doesn't change very often, so I didn't think it needed a separate Docker image - otherwise we'd have to keep track of the two Docker images (or different tags?) and pull from a different one depending on the pipeline.

Maybe the Docker image generation should be a separate pipeline, and the two pipelines just refer to the Docker image?

minibroker/scripts/generate-minibroker-repo.sh
alexjh: Yes. Fixed in 1ea3ce7. This file isn't needed any more, there's a generic one that handles the both the minibroker and the customized helm charts.

minibroker/minibroker-check.yaml
```diff
@@ -0,0 +1,451 @@
...
+  jobs:
+  - generate-mariadb
+  - generate-mongodb
+  - generate-postgres
+  - generate-redis
+  - generate-repo
+
+jobs:
+- name: build-minibroker-image
+  serial: true
```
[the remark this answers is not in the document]
alexjh: Removed it in 8ad075f, it was leftover from testing.

alexjh:
@mook-as I separated the build of the helm client's docker image out into another pipeline `support`.

minibroker/minibroker-check.yaml
alexjh: I think it's useful to have a comment there, it tripped me up as well when I was writing the pipeline.

minibroker/minibroker-master.yaml
```diff
@@ -0,0 +1,272 @@
...
+    <<: *helm-charts-src-configs
+    paths:
+    - stable/redis
+
+- name: src-postgres
+  type: git
+  source:
+    <<: *helm-charts-src-configs
+    paths:
+    - stable/postgresql
```
mook-as: Huh, it's `postgres` in PRs and `postgresql` in `master`?

minibroker/minibroker-master.yaml
```diff
@@ -0,0 +1,272 @@
...
+- name: minibroker
+  jobs:
+  - build-minibroker-image
+  - generate-minibroker-chart
+  - generate-minibroker-repo
+- name: charts
+  jobs:
+  - generate-mariadb
+  - generate-mongodb
+  - generate-postgres
```
mook-as: `generate-postgres` here and `generate-postgresql` in check.

minibroker/minibroker-master.yaml
```diff
@@ -0,0 +1,272 @@
...
+  - generate-redis
+  - generate-repo
+
+jobs:
+- name: build-minibroker-image
+  serial: true
```
mook-as: This still needs serial? (It isn't in check)

minibroker/minibroker-check.yaml
```diff
@@ -0,0 +1,458 @@
...
+      file: src-ci/minibroker/tasks/generate-repo-check.yaml
+      params:
+        MINIBROKER_BUCKET: ((s3-bucket))
+        CHARTS_DIR: ((helm-charts-s3-prefix))/checks/
+    - put: s3.helm-charts-repo-index
+      params:
+        file: 'repo/index.yaml'
+        acl: public-read
+    on_failure:
+      do:
```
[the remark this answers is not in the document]
alexjh: Fixed in 9c808e5

minibroker/minibroker-master.yaml
alexjh: Fixed in 4a85c51

minibroker/minibroker-master.yaml
alexjh: Fixed in 1588e7b.

minibroker/minibroker-master.yaml
alexjh: Fixed in 1c2860f

minibroker/setup-bucket.sh
```diff
@@ -0,0 +1,16 @@
+#!/usr/bin/env sh
+
+set -e
+set -x
+
+s3cmd mb s3://minibroker-helm-charts
+
+aws s3api put-bucket-versioning --bucket minibroker-helm-charts --versioning-configuration Status=Enabled
```
andreas-kupries: Hi @alexjh 

The `minibroker-helm-charts` constant seems to be the same as `s3-bucket` in the `config-production.yaml`, aka `MINIBROKER_BUCKET` in the pipeline config and scripts.

I believe that we should make sure that there cannot be a mismatch ... I.e. provide the config file to use as argument to this script (maybe default to config production), and use `awk` to pull the value of `s3-bucket` out of it. The syntax is simple enough for `awk` to work.

minibroker/setup-bucket.sh
```diff
@@ -0,0 +1,16 @@
...
+
+cat <<EOF > index.yaml
+apiVersion: v1
+entries: ~
+generated: 1970-01-01T00:00:00.000000000Z
+EOF
+
+s3cmd put --acl-public index.yaml s3://minibroker-helm-charts/kubernetes-charts/index.yaml
```
andreas-kupries: S.a. regarding `minibroker-helm-charts`.
Also, any particular reason for using both `aws s3api ...` and `s3cmd ...` ?

alexjh: No good reason. :) I had to use the `aws s3api` to set the versioning because s3cmd didn't support it. Updated to use it everywhere.

784204c

minibroker/setup-bucket.sh
alexjh: Fixed in 784204c

andreas-kupries:
LGTM, and the test pipelines look good also. Merging.

PR #138, opened 2018-07-06
Harts aks deploy improvements

HartS:
Add lb rules for ports 20000-20009 in aks deploy. These allow brain tests to pass.

Also prep AKS cluster for CAP in deploy-aks script

- Create the persistent storageclass
- Install Tiller, and create cluster-admin role and binding
  (The cluster-admin role is no longer defined with newer kubes)
- Apply PodSecurityPolicy workaround
- Create cap-values configmap with IP information


prabalsharma:
LGTM

PR #139, opened 2018-07-06
disable UAA HA for vesrion 2.10.1 and less

prabalsharma:


qa-pipelines/tasks/cf-deploy-upgrade-common.sh
```diff
@@ -127,7 +127,7 @@ set_uaa_sizing_params() {
         if semver_is_gte $(helm_chart_version) 2.11.0; then
             HELM_PARAMS+=(--set=config.HA=true)
         else
-            HELM_PARAMS+=(--set=sizing.HA=true)
+            HELM_PARAMS+=(--set=sizing.HA=false)
```
HartS: You can just take the whole else section out I think. We don't need to explicitly set this to false right? Maybe add a comment saying HA UAA isn't supposed to work prior to 2.11.0

prabalsharma: true

PR #141, opened 2018-07-17
Remove `"` from az cli output for public_ip

HartS:
This is required to not have a literal `"` in the public_ip value inserted into the configmap.. this breaks deployment because domains can't have literal quotation marks.

### DIFF
--- a/qa-tools/deploy-aks.sh
+++ b/qa-tools/deploy-aks.sh
@@ -117,7 +117,7 @@ for i in $CAP_PORTS; do
 done
 
 internal_ips=($(az network nic list --resource-group $AZ_MC_RG_NAME | jq -r '.[].ipConfigurations[].privateIpAddress'))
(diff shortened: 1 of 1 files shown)


qa-tools/deploy-aks.sh
```diff
@@ -117,7 +117,7 @@ for i in $CAP_PORTS; do
 done
 
 internal_ips=($(az network nic list --resource-group $AZ_MC_RG_NAME | jq -r '.[].ipConfigurations[].privateIpAddress'))
-public_ip=$(az network public-ip show --resource-group $AZ_MC_RG_NAME --name $AZ_AKS_NAME-public-ip --query ipAddress)
+public_ip=$(az network public-ip show --resource-group $AZ_MC_RG_NAME --name $AZ_AKS_NAME-public-ip --query ipAddress | sed 's/"//g')
```
mook-as: `tr -d '"'` also works.
Of course, you probably actually want `az … --output tsv` instead.

HartS:
It works as long as you remove the quotes, hence the PR :stuck_out_tongue_winking_eye:

PR #146, opened 2018-07-25
Set garden rootfs driver from cap-values configmap

HartS:
Verified that this sets the right parameter on the helm release when the `garden-rootfs-driver` key is in the `cap-values` configmap

### DIFF
--- a/qa-pipelines/tasks/cf-deploy-upgrade-common.sh
+++ b/qa-pipelines/tasks/cf-deploy-upgrade-common.sh
@@ -120,6 +120,7 @@ set_helm_params() {
     if [ -n "${KUBE_ORGANIZATION:-}" ]; then
         HELM_PARAMS+=(--set "kube.organization=${KUBE_ORGANIZATION}")
     fi
+    HELM_PARAMS+=(--set "env.GARDEN_ROOTFS_DRIVER=${garden_rootfs_driver}")
 }
 
 set_uaa_sizing_params() {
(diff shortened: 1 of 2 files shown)


qa-pipelines/tasks/cf-deploy-upgrade-common.sh
```diff
@@ -120,6 +120,11 @@ set_helm_params() {
     if [ -n "${KUBE_ORGANIZATION:-}" ]; then
         HELM_PARAMS+=(--set "kube.organization=${KUBE_ORGANIZATION}")
     fi
+    if semver_is_gte 2.11.0 $(helm_chart_version); then
```
prabalsharma: I don't this has anything to do with version.
the condition can be: if garden_rootfs_driver==overlay-xfs:
                                   then: --set "env.GARDEN_ROOTFS_DRIVER=overlay-xfs"
we don't need `else conditions` as btrfs is default and we wont set garden_rootfs_driver in cap-values for pb or ecp

qa-pipelines/tasks/cf-deploy-upgrade-common.sh
```diff
@@ -153,7 +158,9 @@ set -o allexport
 # The external_ip is set to the internal ip of a worker node. When running on openstack or azure, 
 # the public IP (used for DOMAIN) will be taken from the floating IP or load balancer IP.
 external_ip=$(kubectl get configmap -n kube-system cap-values -o json | jq -r '.data["internal-ip"]')
-public_ip=$(kubectl get configmap -n kube-system cap-values -o json | jq -r '.data["public-ip"] // empty')
+public_ip=$(kubectl get configmap -n kube-system cap-values -o json | jq -r '.data["public-ip"]')
+garden_rootfs_driver=$(kubectl get configmap -n kube-system cap-values -o json | jq -r '.data["garden-rootfs-driver"] // "btrfs"')
+garden_use_btrfs=$([[ ${garden_rootfs_driver} == btrfs ]] && echo true || echo false)
```
prabalsharma: don't need this

HartS: We still need this to set both variables correctly; with the new commit, we're setting both regardless of the chart version, as having both set won't hurt and will allow our pipeline to continue working with future versions as well as builds from develop (as Andreas noted, support for the new variable is required in those situations).

prabalsharma: https://github.com/SUSE/scf/blob/develop/container-host-files/etc/scf/config/role-manifest.yml#L2670n empty value means BTRFS is used. Set to "false" to use overlay-xfs.

PR #148, opened 2018-07-26
Enable pre upgrade cats for custom scf bundles

HartS:
Integrates https://github.com/SUSE/cf-ci/pull/146 for purposes of testing with AKS clusters.

This adds the acceptance tests to the list of tasks which can be run
'pre-upgrade', but this is intended for pipelines which don't actually
perform an upgrade. In this case, the 'cap-\*-url's in the config-\*.yml
file used can be changed to a custom URL where a zip of the desired
charts can be fetched from, and a preset file which only sets
'-pre-upgrade' flags to true can be used for deploying the pipeline

### DIFF
--- a/qa-pipelines/flags.yml
+++ b/qa-pipelines/flags.yml
@@ -11,8 +11,15 @@ enable-cf-deploy-pre-upgrade: true
 enable-cf-smoke-tests-pre-upgrade: true
 enable-cf-brain-tests-pre-upgrade: true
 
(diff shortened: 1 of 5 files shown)


HartS:
Needs a rebase, at the very least to not integrate work which was removed from the garden rootfs driver branch

HartS:
Cherry-picked the commits we want, and added one more commit to scrub registry passwords which could end up in the pipeline logs.

PR #162, opened 2018-08-23
Remove rbac auth option as it now is the default for a deployment

greygoo:


prabalsharma:
Chris were you able to test this (if removing rbac does not break CI) by deploying CI through this branch?

greygoo:
Yes this was tested by deploying a pipeline with it. A nfs testcase did fail, but that alos failed with master, so I do not think that is related

PR #167, opened 2018-08-28
create pre-upgrade-user and check it after upgrade

prabalsharma:
This will tell us post-upgrade, condition of UAA DB

### DIFF
--- a/qa-pipelines/tasks/cf-upgrade.sh
+++ b/qa-pipelines/tasks/cf-upgrade.sh
@@ -33,11 +33,13 @@ monitor_url() {
   done
 }
 
+# Create pre-upgrade user
 # push app in subshell to avoid changing directory
 (
   cd ci/sample-apps/go-env
   cf api --skip-ssl-validation "https://api.${DOMAIN}"
   cf login -u admin -p changeme -o system
+  cf create-user pre-upgrade-user pre-upgrade-user
   cf create-org testorg
   cf target -o testorg
   cf create-space testspace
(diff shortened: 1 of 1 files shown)


qa-pipelines/tasks/cf-upgrade.sh
```diff
@@ -33,11 +33,13 @@ monitor_url() {
   done
 }
 
+# Create pre-upgrade user
 # push app in subshell to avoid changing directory
 (
   cd ci/sample-apps/go-env
   cf api --skip-ssl-validation "https://api.${DOMAIN}"
   cf login -u admin -p changeme -o system
+  cf create-user pre-upgrade-user pre-upgrade-user
```
HartS: Move this and other lines unrelated to pushing the app out of the subshell for clarity, since we're using that logged-in state elsewhere.

qa-pipelines/tasks/cf-upgrade.sh
```diff
@@ -77,9 +79,11 @@ helm upgrade scf ${CAP_DIRECTORY}/helm/cf${CAP_CHART}/ \
 
 # Wait for CF namespace
 wait_for_namespace "${CF_NAMESPACE}"
-echo "Post Upgrade Orgs State:"
+echo "Post Upgrade Users and Orgs State:"
 cf api --skip-ssl-validation "https://api.${DOMAIN}"
 cf login -u admin -p changeme -o system
+oauth-token=$(cf oauth-token)
```
HartS: we should replace 85-86 with `cf curl`

PR #178, opened 2018-09-24
add capci config for dev upgrades CI, update CAP install urls to 2.13.3

prabalsharma:


qa-pipelines/config-azure.yml
```diff
@@ -26,5 +26,5 @@ organization: splatform
 
 magic-dns-service: omg.howdoi.website
 
-cap-sle-url: https://s3.amazonaws.com/cap-release-archives/master/scf-sle-2.11.0%2Bcf1.36.0.0.g4c9e9ae1.zip
-cap-opensuse-url: https://s3.amazonaws.com/cap-release-archives/master/scf-opensuse-2.11.0%2Bcf1.36.0.0.g4c9e9ae1.zip
+cap-sle-url: https://s3.amazonaws.com/cap-release-archives/master/scf-sle-2.13.3%2Bcf2.7.0.0.gf95d9aed.zip
+cap-opensuse-url: https://s3.amazonaws.com/cap-release-archives/master/scf-opensuse-2.13.3%2Bcf2.7.0.0.gf95d9aed.zip
```
mook-as: Having the URLs in a config (rather than a resource that just grabs the latest release / master, or something) seems odd, but :man_shrugging:

qa-pipelines/config-capci.yml
```diff
@@ -0,0 +1,32 @@
...
+kube-pool-pool: cap-ci-kube-hosts
+
+s3-config-endpoint: ~
+s3-config-access-key: *aws-access-key
+s3-config-secret-key: *aws-secret-key
+s3-config-bucket-opensuse: cap-release-archives
+s3-config-prefix-opensuse: nightly/
+
+s3-config-bucket-sles: cap-release-archives
+s3-config-prefix-sles: nightly/
```
mook-as: Currently I believe we don't produce SLES nightlies

qa-pipelines/config-azure.yml
prabalsharma: Its the specific release from which we need to upgrade from. we don't grab it from resource because all RCs goes in /master/ . We also use this as, kind of a hack for any custom deployments where we need a specific build without doing much change to the code base

qa-pipelines/config-capci.yml
prabalsharma: mm that's ok. we wont turn on that job for now. Ultimately, we need to test sles nightlies though :)

mook-as: Yes,  testing SLES nightlies is good, but we need to have them to test ;)

PR #181, opened 2018-10-03
Wait for release jobs to complete before ns pods

HartS:
This adds an additional wait for job completion (determined from the helm release data) before moving on to waiting for pods to become ready.

### DIFF
--- a/qa-pipelines/tasks/cf-deploy-upgrade-common.sh
+++ b/qa-pipelines/tasks/cf-deploy-upgrade-common.sh
@@ -60,23 +60,46 @@ is_namespace_ready() {
         | sort \
         | uniq) ]]
 }
-
-wait_for_namespace() {
-    local namespace="$1"
+wait_for_jobs() {
+    local release=$1
+    local jobs_desired_remaining
     start=$(date +%s)
-    for (( i = 0  ; i < 960 ; i ++ )) ; do
+    for (( i = 0 ; i < 480 ; i ++ )); do
+        # Get the list of all jobs in the helm release, and subtract the value of 'completed' from 'desired'
+        # It would be better to parse this from `helm get manifest`, but since that command is broken in helm 
+        # v2.8.2 (https://github.com/helm/helm/issues/3833) for now we'll parse it from the human-readable status
+        jobs_desired_remaining=$(helm status $release | awk '/==> v1\/Job/ { getline; getline; while (NF>0) { print $2 - $3; getline } }')
+        now=$(date +%s)
+        if [[ $(echo "${jobs_desired_remaining}" | sort -n | tail -1) -le 0 ]]; then
+           printf "\rDone waiting for %s jobs at %s (%ss)..." "${release}" "$(date --rfc-2822)" $((${now} - ${start}))
+           return 0
+        else
+           sleep 10
+        fi
+        printf "\rWaiting for %s jobs at %s (%ss)..." "${release}" "$(date --rfc-2822)" $((${now} - ${start}))
+    done
+    printf "%s jobs not completed\n" "${release}"
+    return 1
+}
+
+wait_for_release() {
+    local release=$1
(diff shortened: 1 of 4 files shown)


qa-pipelines/tasks/cf-deploy-upgrade-common.sh
```diff
@@ -60,23 +60,46 @@ is_namespace_ready() {
...
+    done
+    printf "%s jobs not completed\n" "${release}"
+    return 1
+}
 
-wait_for_namespace() {
-    local namespace="$1"
+wait_for_release() {
+    local release=$1
+    local namespace=$(helm list | awk '$1=="'"$release"'" {print $NF}')
```
mook-as: `helm list "${release}" | awk '$1=="'"$release"'" { print $NF }'` (to avoid having so many releases that it runs off the bottom)?

qa-pipelines/tasks/cf-deploy-upgrade-common.sh
```diff
@@ -60,23 +60,46 @@ is_namespace_ready() {
...
+        printf "\rWaiting for %s jobs at %s (%ss)..." "${release}" "$(date --rfc-2822)" $((${now} - ${start}))
+    done
+    printf "%s jobs not completed\n" "${release}"
+    return 1
+}
 
-wait_for_namespace() {
-    local namespace="$1"
+wait_for_release() {
+    local release=$1
```
mook-as: (and the other place) technically `$1` should be quoted as `"$1"` because it may contain spaces, but in practice I doubt that's a problem. Would shut up shellcheck though :D

qa-pipelines/tasks/cf-deploy-upgrade-common.sh
```diff
@@ -60,23 +60,46 @@ is_namespace_ready() {
...
 }
+wait_for_jobs() {
+    local release=$1
+    local jobs_desired_remaining
+    start=$(date +%s)
+    for (( i = 0 ; i < 480 ; i ++ )); do
+        # Get the list of all jobs in the helm release, and subtract the value of 'completed' from 'desired'
+        # It would be better to parse this from `helm get manifest`, but since that command is broken in helm 
+        # v2.8.2 (https://github.com/helm/helm/issues/3833) for now we'll parse it from the human-readable status
+        jobs_desired_remaining=$(helm status $release | awk '/==> v1\/Job/ { getline; getline; while (NF>0) { print $2 - $3; getline } }')
```
mook-as: Alternatively, do the counting in awk:
```awk
BEGIN         { j=0; c=0 }
/==> v1\/Job/ { getline ; getline ; j=1 }
NF == 0       { j=0 }
j > 0         { c += $2 - $3 }
END           { print c }
```

qa-pipelines/tasks/cf-deploy-upgrade-common.sh
```diff
@@ -60,23 +60,46 @@ is_namespace_ready() {
...
+        now=$(date +%s)
+        if [[ $(echo "${jobs_desired_remaining}" | sort -n | tail -1) -le 0 ]]; then
```
mook-as: Shouldn't that be `-gt`? You have (desired - successful), so it should never be negative.

mook-as: Oh, because I'm an idiot and read the condition backwards; `-le 0` for accept.

qa-pipelines/tasks/cf-deploy-upgrade-common.sh
mook-as: Ah, apparently it's possible to have completed > desired in some situations (kube just ensures that at least one succeeds). So we can't do what I have (unless we clamp `$2 - $3` to `0`), so taking the loop as-is with the `sort` is fine.

PR #184, opened 2018-10-05
Workaround for brain pod containing more objects

HartS:
Now that the acceptance-tests-brain.yaml contains some additional
objects necessary to grant the right permissions to the brain tests, the
run-test script has to create those separately, before running just the
pod yaml with `kubectl run`. Fortunately nothing in the additional
object specs requires substitution, so creating these is pretty
straightforward

### DIFF
--- a/qa-pipelines/tasks/run-test.sh
+++ b/qa-pipelines/tasks/run-test.sh
@@ -28,8 +28,8 @@ set -o allexport
 EXCLUDE_BRAINS_PREFIX=011
 # Set this to define number of parallel ginkgo nodes in the acceptance test pod
 ACCEPTANCE_TEST_NODES=3
-DOMAIN=$(kubectl get pods -o json --namespace scf api-0 | jq -r '.spec.containers[0].env[] | select(.name == "DOMAIN").value')
 CF_NAMESPACE=scf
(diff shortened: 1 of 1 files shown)


qa-pipelines/tasks/run-test.sh
```diff
@@ -74,18 +75,38 @@ EOF
...
 container_status() {
-  kubectl get --output=json --namespace=scf pod $1 \
+  kubectl get --output=json --namespace="${CF_NAMESPACE}" pod $1 \
     | jq '.status.containerStatuses[0].state.terminated.exitCode | tonumber' 2>/dev/null
 }
 
 image=$(awk '$1 == "image:" { gsub(/"/, "", $2); print $2 }' "${CAP_DIRECTORY}/kube/cf${CAP_CHART}/bosh-task/${TEST_NAME}.yaml")
 
+test_pod_yml="${CAP_DIRECTORY}/kube/cf${CAP_CHART}/bosh-task/${TEST_NAME}.yaml"
+test_non_pods_yml=
```
prabalsharma: why this, if its defined below?

prabalsharma:
Did you test this with upgrades? Will brains work with 2.13.3 and also 2.14.1?

qa-pipelines/tasks/run-test.sh
HartS: Because nounset throws an error for the other tests if it's not set

PR #186, opened 2018-10-05
Deploy Azure broker in cf-deploy from configmap info

HartS:
I haven't had the opportunity to test this yet since some recent minor changes/fixes.


HartS:
Made another commit after testing revealed some additional bugs in the deploy-aks script, specifically where it prepares some stuff via helm and sets the redis hostname in the configmap for the azure cf broker deployment to use. I'm pretty confident this will work for a deployment of CF after installing AKS now. However, it still remains to be seen whether things need to be cleaned in the Redis datastore between CF deployments (in the process of testing)

prabalsharma:
this should be part of qa-tests

PR #194, opened 2018-11-15
Use new api role name deploys+upgrades w SCALED_HA

HartS:


prabalsharma:
Looks invalid now. api_role_name=api in CAP 1.3.1 is not needed. Also, CAP 1.3.1 will no longer be supported by QA CI as CAP 1.4 is out.
This can be handled in https://trello.com/c/RxYjkfkr/661-ci-fix-scaledha

PR #198, opened 2018-11-30
Use Standard_D4_v2 for AKS

HartS:
We're seeing failures with HA deployments that have credhub enabled with
any smaller instance types


f0rmiga:
Closing as it is not needed for now. Let's open another PR if the current instance type doesn't fit in the future.

PR #200, opened 2018-12-07
Add GitHub status on all pipelines

mook-as:
This will add a GitHub status marker on the commit associated with the SCF input archive.  The name of the status context is derived from the pipeline name.


mook-as:
Marking WIP for now because I still need @colstrom to update https://hub.docker.com/r/resource/github-status/

mook-as:
Concourse resource updated, this can now be deployed.

PR #202, opened 2019-01-10
Clear legacy PSPs/CRs/CRBs and create new ones

HartS:
This PR supersedes https://github.com/SUSE/cf-ci/pull/199

```
- Create privileged and unprivileged PSPs
- Link to both in deploy parameters for CAP >= 2.14.5
- Create legacy privileged CRB for default SAs in CAP < 2.14.5
```

Changes on the new branch include:
- Based on current master with new notification changes (work from previous branch, `harts-use-privileged-and-unprivileged-psp` has been cherry-picked)
- Upon discussion with @bikramnehra renamed cap-psp-unprivileged.yaml -> cap-psp-nonprivileged.yaml, the psp name cap.psp.unprivileged -> cap.psp.nonprivileged (and updated references in other files accordingly). `unprivileged` sounds more natural to me, but since the helm param to CAP is `nonprivileged` we should stick with that for consistency.
- Fixed the accidental inversion of the psp names when passing them to helm params (which caused failures in previous tests)

### DIFF
--- a/automation-scripts/prep-new-cluster.sh
+++ b/automation-scripts/prep-new-cluster.sh
@@ -58,9 +58,6 @@ if ! kubectl get storageclass persistent &> /dev/null; then
   kubectl create -f "$(dirname $0)/nfs-provisioner"
 fi
 
-echo "Enable PodSecurityPolicy"
-kubectl create -f "$(dirname $0)/../qa-tools/cap-psp-rbac.yaml"
-
 echo "Ensure the following config contents are in the lockfile for your concourse pool kube resource:"
 echo "---"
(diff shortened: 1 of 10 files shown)


HartS:
@satadruroy 

> Couple of comments
> 
> * Can we pass in a k8s version in the deploy-aks.sh (`az aks create -k...`) and default it to 1.11.5 maybe? Without an explicit version AKS picks 1.9.1 which is quite old and has CVE-2018-1002105.

I agree specifying the kube version to deploy will help with consistency, as the default deployed by AKS changes regularly. However, that should probably be a separate PR (it's out of the scope of this branch)

> * As previously discussed on RC, do we really need the `cap-cr-privileged`? The CRs and CRBs should be created automatically once we pass in the PSPs in the helm chart values.

I took another look, and since the clusterrole created here is referenced from the bindings created for tests to pass, it makes sense to leave it for now. As we start testing the next release, I think we should create the bindings to the CAP-created roles (and at that point we can delete the PSPs defined here as well, since CAP will create those by default too)

PR #204, opened 2019-01-17
Allow both EXCLUDE and INCLUDE to be set for brains tests

HartS:


HartS:
Closing due to work being included with resolutions in PR #240

PR #205, opened 2019-01-21
Image diff script

greygoo:
This adds a script takes two directories containing different versions of uaa and scf helm charts, gets the image sources and retrieves the image blobSums from the docker repos. It then contains the blobSums to make sure we don't have identical images in two different  builds


greygoo:
New PR with only the script and without the concourse integration: https://github.com/SUSE/cf-ci/pull/216

HartS:
@greygoo Just to be clear, this PR should be closed now right? I think #216 superceded this and contained the relevant work

PR #211, opened 2019-01-25
Add minibroker-redis brain test to exclude pattern

HartS:
This test is failing currently due to a client/server helm version
mismatch


HartS:
This is still failing (along with some other tests now), but https://github.com/SUSE/cf-ci/pull/218 adds all the failing tests to the excludes. We can revert it when we get to the next RC

PR #212, opened 2019-02-09
Delete predefined PSPs for CAP >=2.15.1

HartS:
This verifies that charts now create necessary PSP defaults


prabalsharma:
filename ci/qa-tools/cap-cr-privileged-2.15.1.yaml missing

PR #214, opened 2019-02-14
Let CAP charts manage LB for AKS deploys

HartS:


qa-pipelines/tasks/cf-deploy.yml
```diff
@@ -4,6 +4,7 @@ image_resource:
   type: docker-image
   source:
     repository: splatform/cf-ci-orchestration
+    tag: services-loadbalanced
```
prabalsharma: we should use latest now. Since, we have already tested this

qa-pipelines/tasks/cf-teardown.yml
```diff
@@ -4,6 +4,7 @@ image_resource:
   type: docker-image
   source:
     repository: splatform/cf-ci-orchestration
+    tag: services-loadbalanced
```
prabalsharma: Don't forget to push the docker image to latest when you remove this

cf-ci-orchestration/Dockerfile
```diff
@@ -19,6 +20,10 @@ RUN true \
     tar \
     direnv \
     bash-completion docker-bash-completion systemd-bash-completion \
+    python-xml python-pyOpenSSL python-devel \
```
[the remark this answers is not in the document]
HartS: we can't put inline comments on these run-on commands. Also, nothing else is being commented here. These are prerequisites for the az client

PR #216, opened 2019-02-18
Add script to compare image checksums from two helm charts

greygoo:
This PR replaces https://github.com/SUSE/cf-ci/pull/205
It leaves away the concourse integration and does only add the script which can be used manually


svollath:
LGTM

PR #218, opened 2019-02-20
Skip failing brain tests:

HartS:
- 010 insecure_registry
- 011 nfspersi
- 012 minibroker_redis


HartS:
`minibroker_redis` may only be failing now on platforms using managed loadbalancers because some dns settings may be required. I don't see how we'd be able to set those appropriately from the brain tests though, without using platform-specific credentials baked into a secret (which QA is using already)

prabalsharma:
this looks invalid now. All brains should pass for AKS, EKS

PR #224, opened 2019-03-15
GKE cluster automation with cf-ci container

satadruroy:


satadruroy:
There are couple of issues with the PR:

- swap accounting needs to be set in a different way to persist across node image upgrades.  
- The cf-ci container is getting nuked when the script exits - need to look into it.

cf-ci-orchestration/Dockerfile
```diff
@@ -26,6 +26,7 @@ RUN true \
     && rpm --import https://packages.microsoft.com/keys/microsoft.asc \
     && zypper addrepo --name 'Azure CLI' --check https://packages.microsoft.com/yumrepos/azure-cli azure-cli \
     && zypper --non-interactive install --from azure-cli azure-cli \
+    && zypper refresh \
```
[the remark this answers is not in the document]
satadruroy: yeah, the `zypper refresh` was an accidental left-over from the time I was trying to do the glcoud install via zypper - did not work out but I forgot to take the refresh out.

qa-tools/deploy-gke.sh
```diff
@@ -0,0 +1,127 @@
...
+postfix=$(cat /dev/urandom | env LC_CTYPE=C tr -dc 'a-z0-9' | fold -w 16 | head -n 1)
+export CLUSTER_NAME=${CLUSTER_NAME-cap-${postfix}}
+export CLUSTER_ZONE=${CLUSTER_ZONE-us-west1-a}
+
+#  We are setting up a zonal cluster, not a regional one  
+
+export NODE_COUNT=3
+export SA_USER=$1
+export KEY_FILE=$2
+export PROJECT=${3-suse-css-platform}
```
mook-as: Did you mean to say `${3:-suse-css-platform}` (and similarly for `CLUSTER_NAME` and `CLUSTER_ZONE`).

The difference seems to be that if you run this as `deploy-gke.sh foo bar ''` (where `$3` is explicitly empty), `${3:-}` will use the default but `${3-}` will not.

HartS:
I also need to `export CLOUDSDK_PYTHON_SITEPACKAGES=1` for gcloud to work

qa-tools/deploy-gke.sh
```diff
@@ -0,0 +1,127 @@
...
+# Log out from any existing context and log in with the GKE service account
+
+gcloud auth revoke
```
HartS: This should have `|| true` because it fails when there's no existing gcloud auth.

qa-tools/deploy-gke.sh
```diff
@@ -0,0 +1,127 @@
+usage() {
+  cat << 'EOF'
+  Usage:
+  deploy-gke.sh [SERVICE_ACCOUNT_NAME] [KEY_FILE] [PROJ_ID] (optional)
```
HartS: `SERVICE_ACCOUNT_EMAIL` might be more clear than `SERVICE_ACCOUNT_NAME`

qa-tools/deploy-gke.sh
satadruroy: yes, oversight...

qa-tools/deploy-gke.sh
```diff
@@ -0,0 +1,127 @@
...
+gcloud auth revoke
+gcloud auth activate-service-account $SA_USER --key-file $KEY_FILE --project=$PROJECT
+
+# Clusters on k8s 1.10 and above will no longer get compute-rw and storage-ro scopes
+gcloud config set container/new_scopes_behavior true
+
+# Create GKE cluster 
+gcloud container clusters create ${CLUSTER_NAME} --image-type=UBUNTU --machine-type=n1-standard-4 --zone \
+${CLUSTER_ZONE} --num-nodes=$NODE_COUNT --no-enable-basic-auth --no-issue-client-certificate \
+--no-enable-autoupgrade --metadata disable-legacy-endpoints=true
```
mook-as: It would be nice to set labels for the owner, to make it easier to track. Something like:

```bash
--labels=owner=$(gcloud config get-value account | tr [:upper:] [:lower:] | tr -c a-z0-9_- _ )
```

If this is for CI, it might be easier to hard code the owner (but for manually-run QA instances that wouldn't work as well).

PR #227, opened 2019-03-25
set INSECURE_DOCKER_REGISTRIES

prabalsharma:


qa-pipelines/tasks/lib/cf-deploy-upgrade-common.sh
```diff
@@ -215,6 +215,9 @@ else
     DOMAIN=${public_ip}.${MAGIC_DNS_SERVICE}
 fi
 
+#Set INSECURE_DOCKER_REGISTRIES for brain test
+INSECURE_DOCKER_REGISTRIES=\"insecure-registry.tcp.${DOMAIN}:20005\"
```
f0rmiga: Why escaped double quotes?

prabalsharma: it fails otherwise

f0rmiga: Well, then it's nicer to handle the escaped double quotes where `INSECURE_DOCKER_REGISTRIES` is used:
```
--set "env.INSECURE_DOCKER_REGISTRIES=\"${INSECURE_DOCKER_REGISTRIES}\""
```

PR #229, opened 2019-04-02
Improvements for GKE deployment

bikramnehra:


qa-tools/destroy-gke.sh
```diff
@@ -0,0 +1,19 @@
...
+	usage
+	exit 0
+fi
+
+export CLUSTER_NAME=$1
+
+gcloud -q container clusters delete $CLUSTER_NAME
+
+# Pruning all the PV disks attached to the cluster
+gcloud -q compute disks delete $(gcloud compute disks list --filter="name~'$CLUSTER_NAME'")
```
satadruroy: It's not just the disks, some other resources such as allocated IP addresses are sometimes not cleaned up. Without going through the entire graph of dependent resources it's hard to know which ones are not cleaned up.

HartS: We should make a card to ensure the teardown script cleans up all resources then

PR #230, opened 2019-04-08
Enable app autoscaler

prabalsharma:


qa-pipelines/qa-pipeline.yml
```diff
@@ -95,6 +95,7 @@ jobs:
         KUBE_ORGANIZATION: ((organization))
         HA: false
         SCALED_HA: false
+        AutoScaler: true
```
HartS: Why are we mixing cases here? This should be all caps like the rest of them

prabalsharma: Because its not an abbreviation. If you really want it, I guess I can make the change, but this looked correct to me.

prabalsharma: @HartS you want APP_AUTO_SCALER or AUTOSCALER?

PR #233, opened 2019-04-09
Add --pr flag to set-pipeline to test PR build

HartS:


qa-pipelines/set-pipeline
```diff
@@ -17,31 +17,35 @@ Available options:
     -t, --target=       Set fly target concourse
     -b, --branch=       Use a specific branch or tag. Defaults to the current branch
     --nightly           Deploy pipeline that triggers when new nightly builds are released
+    --pr=               Deploy pipeline that tests a build from a specific PR
```
prabalsharma: why there is a `=` ?

qa-pipelines/tasks/lib/cf-deploy-upgrade-common.sh
```diff
@@ -21,7 +21,7 @@ if [ -n "${CAP_INSTALL_VERSION:-}" ]; then
     export CAP_DIRECTORY=cap-install-version
     unzip ${CAP_DIRECTORY}.zip -d ${CAP_DIRECTORY}/
 else
-    unzip ${CAP_DIRECTORY}/scf-*.zip -d ${CAP_DIRECTORY}/
+    unzip ${CAP_DIRECTORY}/*scf-*.zip -d ${CAP_DIRECTORY}/
```
prabalsharma: why `*`?

qa-pipelines/set-pipeline
prabalsharma: ok, because it takes the PR value `s3_config_prefix=prs/PR-${2}-`
Please add usage in Readme along with an example.

qa-pipelines/tasks/lib/cf-deploy-upgrade-common.sh
HartS: @prabalsharma because the name of this for pr builds will be `pr-xxxx-scf-...`

HartS: where `xxxx` is the number of the PR (and also the parameter passed to the set-pipeline command)

PR #234, opened 2019-04-11
adding script for retrieving cluster resource metrics

bikramnehra:


HartS:
What do you think about defining a `print_with_label` function that takes the label name as the first parameter, and the quoted message as parameter 2? Since it looks like all the labels are being printed with CYAN and NC character escapes, it would make the output that gets labels a little bit more readable.

bikramnehra:
added new PR: https://github.com/SUSE/cf-ci/pull/235

PR #239, opened 2019-04-26
Update ns readiness check

HartS:


HartS:
This branch is based on the gke https://github.com/SUSE/cf-ci/tree/harts-deploy-cap-on-gke-eks , so should not be merged before https://github.com/SUSE/cf-ci/pull/238

PR #240, opened 2019-04-27
Run single brain test

HartS:
Deploy a pipeline which presents each brain test as a separate job (and omits all other tasks), in order to run an individual brain test.


qa-pipelines/pipeline-presets/single-brain.yml
```diff
@@ -0,0 +1 @@
+---
```
bikramnehra: is this needed?

qa-pipelines/tasks/run-test.sh
```diff
@@ -23,7 +23,9 @@ fi
 set -o nounset
 set -o allexport
 # Set this to skip a test, e.g. 011
-EXCLUDE_BRAINS_PREFIX=''
+EXCLUDE_BRAINS_PREFIX=
```
bikramnehra: Wasn't " " better way to handle the default case?

qa-pipelines/tasks/run-test.sh
```diff
@@ -23,7 +23,9 @@ fi
...
+# Set this to run only one test. If EXCLUDE and INCLUDE are both specified, EXCLUDE is applied after INCLUDE
+INCLUDE_BRAINS_PREFIX=${INCLUDE_BRAINS_PREFIX:-}
```
bikramnehra: Can we be more clear in comments about the expected/acceptable values for EXCLUDE_BRAINS_PREFIX and INCLUDE_BRAINS_PREFIX?

qa-pipelines/pipeline-presets/single-brain.yml
HartS: It's good practice to begin yml files to start with `---` which is the yml document separator (it allows multiple yml documents to exist in one file or stream). The QA pipeline deployment file concatenates several ymls when deploying pipelines (though I think just for values). It might not be necessary here (because I don't think other ymls get concatenated with the pipeline definition) but it does make it more future-proof.

qa-pipelines/tasks/run-test.sh
HartS: It's useful to leave this here, with a blank value, because we change this when there are, for example, failing brains tests with a build, or new tests which we expect to skip. Without a value, nothing is excluded, but it does provide a stub and a clear place for us to modify this later if necessary

PR #244, opened 2019-05-13
Adding minibroker chart extractor

bikramnehra:
## Description

Script to extract versions of minibroker databases known to be working with CAP.

**Details:** https://trello.com/c/iIrYZZY7/1068-5-curate-minibroker-service-plans-that-work-with-cap

## Test plan

Run `python extractor.py <path/to/versions.yaml> <path/to/output/index.yaml>` and check the generated `index.yaml`

### DIFF
--- a/automation-scripts/minibroker-charts/extractor.py
+++ b/automation-scripts/minibroker-charts/extractor.py
@@ -1,54 +1,72 @@
 #!/usr/bin/env python
 
 # Script to extract working versions of minibroker database charts from helm stable repo's index.yaml.
-# Make sure the YAML parsing lib is installed: `pip install ruamel.yaml`.
+# Make sure the ruamel.yaml and requests is installed: `pip install ruamel.yaml/requests`.
 # 
 # Usage:
 # 1. Update the versions.yaml with the minibroker DB versions known to be working with current CAP release.
-# 2. Run `python extractor.py`
+# 2. Run `python extractor.py <path/to/versions.yaml> <path/to/output/index.yaml>`
+#
 
-import os
+import requests
+from sys import argv
 from ruamel.yaml import YAML
 
-os.system("curl -s https://kubernetes-charts.storage.googleapis.com/index.yaml > helm_stable_index.yaml")
+HELM_REPO_URL="https://kubernetes-charts.storage.googleapis.com/index.yaml"
 
-def read_yaml(file):
+def process_needles():
+    """
+    Method to load and process known working versions in versions.yaml.
+    """
     try:
-        with open(file, 'r') as stream:
-            yaml = YAML()
+        with open(argv[1], 'r') as stream:
+            yaml = YAML(typ='safe')
             return yaml.load(stream)
     except yaml.YAMLError as exc:
         print(exc)
 
-all_versions = read_yaml("helm_stable_index.yaml")
+def process_haystack(databases):
+    """
+    Method to load and process helm stable repo index.yaml.
+    """
+    req = requests.get(HELM_REPO_URL)
+    yaml = YAML()
+    data = yaml.load(req.content)
+    haystack = {}
+    
+    for db in databases:
+        for item in data["entries"][db]:
+            if "appVersion" in item and "version" in item:
+                key = db+item["appVersion"]+item["version"]
+                haystack[key] = item
+
+    return haystack
 
 def dump_yaml(data):
-    with open('index.yaml', 'w') as outfile:
+    """
+    Method to dump output index.yaml file.
+    """
+    with open(argv[2], 'w') as outfile:
         yaml = YAML()
-        data = {"apiVersion": all_versions["apiVersion"], "entries": data}
         yaml.dump(data, outfile)
 
-def exists(db_working_versions, db_version):
-    for dwv in db_working_versions:
-        if "version" in db_version and "appVersion" in db_version:
-            if dwv["appVersion"] == db_version["appVersion"] \
-                and dwv["version"] == db_version["version"]:
-                    return True
-
 def main():
-    entries = {}
-    working_versions = read_yaml("versions.yaml")
+    needles = process_needles()
+    databases = needles.keys()
+    haystack = process_haystack(databases)
+    output = {}
+    for db in databases:
+        for item in needles[db]:
+            needle = db+item["appVersion"]+item["version"]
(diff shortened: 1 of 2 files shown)


automation-scripts/minibroker-charts/extractor.py
```diff
@@ -0,0 +1,54 @@
...
+# Make sure the YAML parsing lib is installed: `pip install ruamel.yaml`.
+# 
+# Usage:
+# 1. Update the versions.yaml with the minibroker DB versions known to be working with current CAP release.
+# 2. Run `python extractor.py`
+
+import os
+from ruamel.yaml import YAML
+
+os.system("curl -s https://kubernetes-charts.storage.googleapis.com/index.yaml > helm_stable_index.yaml")
```
andreas-kupries: - Has python a http package which could be used over spawning a sub process ?
- Use a temp file for the result.
- Move into `main`

automation-scripts/minibroker-charts/extractor.py
```diff
@@ -0,0 +1,54 @@
...
+
+def read_yaml(file):
+    try:
+        with open(file, 'r') as stream:
+            yaml = YAML()
+            return yaml.load(stream)
+    except yaml.YAMLError as exc:
+        print(exc)
+
+all_versions = read_yaml("helm_stable_index.yaml")
```
andreas-kupries: Move into main.

automation-scripts/minibroker-charts/extractor.py
```diff
@@ -0,0 +1,54 @@
...
+
+def dump_yaml(data):
+    with open('index.yaml', 'w') as outfile:
```
andreas-kupries: Make path of destination file configurable via script argument.

automation-scripts/minibroker-charts/extractor.py
```diff
@@ -0,0 +1,54 @@
...
+def exists(db_working_versions, db_version):
+    for dwv in db_working_versions:
+        if "version" in db_version and "appVersion" in db_version:
+            if dwv["appVersion"] == db_version["appVersion"] \
+                and dwv["version"] == db_version["version"]:
+                    return True
+
+def main():
+    entries = {}
+    working_versions = read_yaml("versions.yaml")
```
andreas-kupries: Make path of configuration file configurable via script argument.

automation-scripts/minibroker-charts/extractor.py
```diff
@@ -0,0 +1,54 @@
...
+
+all_versions = read_yaml("helm_stable_index.yaml")
+
+def dump_yaml(data):
+    with open('index.yaml', 'w') as outfile:
+        yaml = YAML()
+        data = {"apiVersion": all_versions["apiVersion"], "entries": data}
+        yaml.dump(data, outfile)
+
+def exists(db_working_versions, db_version):
```
andreas-kupries: What is the needle we are looking for, what is the haystack ?
(Better argument names needed (needle, and haystack ;) ).
(`db_working_versions` seems to be the hackstack, and `db_version` the needle)

Haystack seems to be an array.
Better make it a hash (python: dictionary ?).
Enables check for existence in one step instead of iterating over the array (linear search).

automation-scripts/minibroker-charts/extractor.py
```diff
@@ -0,0 +1,54 @@
...
+                and dwv["version"] == db_version["version"]:
+                    return True
+
+def main():
+    entries = {}
+    working_versions = read_yaml("versions.yaml")
+
+    for db in working_versions:
+        temp_list = []
+        db_working_versions = working_versions[db]
```
andreas-kupries: As the haystack, convert to hash/dictionary for quicker existence checking.

automation-scripts/minibroker-charts/extractor.py
andreas-kupries: Might not even require a helper method/function anymore.

automation-scripts/minibroker-charts/extractor.py
```diff
@@ -0,0 +1,72 @@
...
+    """
+    req = requests.get(HELM_REPO_URL)
+    yaml = YAML()
+    data = yaml.load(req.content)
+    haystack = {}
+    
+    for db in databases:
+        for item in data["entries"][db]:
+            if "appVersion" in item and "version" in item:
+                key = db+item["appVersion"]+item["version"]
```
andreas-kupries: Add separator characters between the parts of the key.

Examples
- appV = 1.0.22, ver=1.4
- appV = 1.0.2, ver=21.4

Without a character keeping appversion and version apart these two result in the same key.

automation-scripts/minibroker-charts/extractor.py
```diff
@@ -0,0 +1,72 @@
...
+
+def dump_yaml(data):
+    """
+    Method to dump output index.yaml file.
+    """
+    with open(argv[2], 'w') as outfile:
+        yaml = YAML()
+        yaml.dump(data, outfile)
+
+def main():
```
andreas-kupries: Add check that argv[1], argv[2] are actually specified, error with usage information when not.
Right now forgetting to set the two arguments will result in python stack traces.

automation-scripts/minibroker-charts/extractor.py
```diff
@@ -0,0 +1,72 @@
...
+    needles = process_needles()
+    databases = needles.keys()
+    haystack = process_haystack(databases)
+    output = {}
+    for db in databases:
+        for item in needles[db]:
+            needle = db+item["appVersion"]+item["version"]
```
andreas-kupries: See above about the issue with/out a separating character.

Two uses of the same code -> place into a helper function which constructs the key from its arguments.

automation-scripts/minibroker-charts/extractor.py
```diff
@@ -0,0 +1,82 @@
...
+    """
+    with open(argv[2], 'w') as outfile:
+        yaml = YAML()
+        yaml.dump(data, outfile)
+
+def main():
+    if len(argv) < 3:
+        print("Please provide correct arguments...")
+        print("Usage: python extractor.py <path/to/versions.yaml> <path/to/output/index.yaml>")
+    else:
```
andreas-kupries: Might be nicer to `exit` here, and leave the `else:` branch unindented.
Not a blocker tough.

PR #246, opened 2019-05-16
fix deploy-aks, mac compatible

prabalsharma:
- mac os compatible
- remove cluster-admin.yaml, as its no longer required


prabalsharma:
won't support mac

PR #248, opened 2019-05-22
Parameterize the CF pipelines

mook-as:
We should only deploy the steps that are part of the pipeline being used, rather than deploying all the steps all the time and checking within each step whether it should continue.

This does _not_ do major refactoring of the pipeline that is needed.


qa-pipelines/set-pipeline
```diff
@@ -87,6 +87,11 @@ if ! test -r "${pipeline_file}" ; then
     exit 1
 fi
 
+alt_secrets_file="CONCOURSE_SECRETS_FILE_${pool//-/_}"
```
HartS: The pool name will be something like "azure", "gke", etc. so the substitution here is unnecessary

I'm guessing you were allowing the secrets to be pulled from a file besides blackbox for debugging? But I'm not really sure what's happening here or if we need it. If we do, a comment and note in the relevant README section about alternate usages would be helpful.

qa-pipelines/flags.yml
```diff
@@ -4,37 +4,45 @@
 # The settings in this file will also be valid for our most complete pipeline
 # (which performs an upgrade and all tests)
 
-# deploy cf for upgrade pipelines
-enable-cf-deploy-pre-upgrade: true
+# Enable HA and non-HA pipelines
```
HartS: Since we're modifying the way flags.yml is used, we need to update this section of the README: https://github.com/SUSE/cf-ci/tree/master/qa-pipelines#preset-file-instructions

Also, think setting the pre-upgrade cats flag to true here affects the functionality of the canonical pipeline, which can be deployed using cap-qa-full-upgrades.yml preset file, as mentioned in that same README section

qa-pipelines/flags.yml
```diff
@@ -4,37 +4,45 @@
...
+sa: true
+
+# Report status as steps complete
+status-reporting: true
+
+# version to upgrade from; if null, upgrade is not performed.
+upgrade-from-version: scf-sle-2.15.2+cf3.6.0.0.gde1bd02f
+
+# Deploy CF USB before upgrading
+enable-cf-usb-deploy-pre-upgrade: true
```
HartS: The deploy USB task and post-upgrade USB task should only be used before and after the upgrade, in upgrade pipelines; it doesn't hurt to clarify that this is a pre-upgrade task, but the post-upgrade component of it should still have `post-upgrade` in the name (in contrast with other tasks which may be run as post-upgrade tasks), because this one should *only* run after an upgrade (where the pre-upgrade USB task ran prior to the upgrade)

qa-pipelines/qa-pipeline.yml.erb
```diff
@@ -0,0 +1,268 @@
...
+<% if status_reporting %>
+- name: status.src
+  type: github-status
+  source:
+    repo: ((src-repo))
+    access_token: ((github-access-token))
+<% end %>
+
+jobs:
+<% %w(SA HA).each do |avail| %>
```
HartS: Looping like this is great, as it allows us to halve the template size, but it's unclear to me why we'd want to deploy pipelines which don't have both jobs already (or pipelines which have no jobs if neither of the flags are set), which can happen by being able to set the ha and sa flags individually

qa-pipelines/set-pipeline
```diff
@@ -155,7 +160,7 @@ fly \
     --non-interactive \
     --pipeline="${pipeline_name}" \
     --var pipeline-name="${pipeline_name}" \
-    --config="${pipeline_file}" \
+    --config=<(ruby generate.rb "${preset_file}" flags.yml) \
```
HartS: Note that there's a separate manifest for the single-brain pipeline (see merged PR #240). Right now generate.rb doesn't account for this and will break when set-pipeline is run with the --single-brain flag. If the flag is set, generate.rb should generate a pipeline with one job for each brain test (see the block which is repeated with minor variations here )

Maybe this could share the template with the regular pipeline, but it might be easier to create a separate template for it.

qa-pipelines/qa-pipeline.yml.erb
```diff
@@ -0,0 +1,268 @@
...
+        state: pending
+    <% end %> # status_reporting
+    on_failure:
+      put: pool.kube-hosts
+      params: {release: pool.kube-hosts}
+  - do:
+    <% if upgrade_from_version %>
+    - task: cf-deploy-pre-upgrade
+      file: ci/qa-pipelines/tasks/cf-deploy.yml
+      params:
```
HartS: I think it'd be cleaner to generate params sections can be constructed by a helper, since a lot of them are repeated (cf-deploy and cf-upgrade for example are the same, `CAP_BUNDLE_URL` and `GKE_PRIVATE_KEY_BASE64` are repeated, and the `TEST_RUN` param can be determined by the task name... the `cf-smoke-tests` task can be renamed `smoke-tests` if that makes it easier

qa-pipelines/tasks/run-test.sh
```diff
@@ -1,16 +1,7 @@
...
-     [[ $ENABLE_CF_SMOKE_TESTS == true ]]; then
-    TEST_NAME=smoke-tests
-elif [[ $ENABLE_CF_BRAIN_TESTS_PRE_UPGRADE == true ]] || \
-     [[ $ENABLE_CF_BRAIN_TESTS == true ]]; then
-    TEST_NAME=acceptance-tests-brain
-elif [[ $ENABLE_CF_ACCEPTANCE_TESTS == true ]] || \
-     [[ $ENABLE_CF_ACCEPTANCE_TESTS_PRE_UPGRADE == true ]]; then
-    TEST_NAME=acceptance-tests
-else
+if [[ -z "${TEST_NAME:-}" ]] ; then
```
HartS: I love that we can delete the whole above mess now. Is this section strictly necessary though?

qa-pipelines/tasks/run-test.yml
```diff
@@ -8,5 +8,8 @@ inputs:
 - name: ci
 - name: s3.scf-config
 - name: pool.kube-hosts
+params:
```
HartS: I'm wondering if it's a good idea to include these default params sections. If a pipeline is deployed without these values set from the config to template generation handoff, we want the pipeline deployment to fail, rather than having the concourse build fail after it gets run

qa-pipelines/qa-pipeline.yml.erb
```diff
@@ -0,0 +1,268 @@
...
+    - task: cf-upgrade
+      file: ci/qa-pipelines/tasks/cf-upgrade.yml
+      params:
+        KUBE_REGISTRY_HOSTNAME: ((registry-hostname))
+        KUBE_REGISTRY_USERNAME: ((registry-username))
+        KUBE_REGISTRY_PASSWORD: ((registry-password))
+        KUBE_ORGANIZATION: ((organization))
+        HA: <%= avail == 'HA' %>
+        MAGIC_DNS_SERVICE: ((magic-dns-service))
+        CAP_BUNDLE_URL: <%= bundle_url_from_version(enable_cf_deploy) %>
```
HartS: `<%= bundle_url_from_version(enable_cf_deploy) %>` just becomes an empty string right? Why don't we just use an empty string here then?

qa-pipelines/qa-pipeline.yml.erb
```diff
@@ -0,0 +1,268 @@
...
+        context: ((pipeline-name))-<%= avail %>
+        description: "QA Pipeline: ((pipeline-name)) (SLES <%= avail %>)"
+        path: commit-id/sha
+        state: pending
+    <% end %> # status_reporting
+    on_failure:
+      put: pool.kube-hosts
+      params: {release: pool.kube-hosts}
+  - do:
+    <% if upgrade_from_version %>
```
HartS: Can you add an inline comment clarifying that we populate this part of the template for upgrade pipelines? But actually, I'm not sure where this is supposed to be set. In our existing config files, we use the parameter `cap-sle-url`

qa-pipelines/flags.yml
HartS: Oops, not sure when the symlink got deleted, but it shouldn't have been.

The flags file *is* a list of flags, with the default settings for the 'canonical' upgrade pipeline. As set-pipeline currently works, it still takes the list of all flags from that file, but sets any that aren't in the passed preset file to false. As it is, only 2 flags are false by default (the non-upgrade deploy shouldn't happen after an upgrade), because of the way concourse builds still had to run through all builds.

Honestly I think this was an incredibly hacky solution which was necessary at the time, but now that we're generating the pipeline manifest and only including the tasks which need to be run, I think we should move some of this logic into the generation script. Maybe we don't even need preset files anymore in their current form, though this could also part of a broader refactor that doesn't necessarily need to happen right now

qa-pipelines/flags.yml
HartS: After looking at this, I don't think USB deploy needs to be two separate tasks. There's a pre-upgrade and post-upgrade component we created as an integration test to reproduce and test resolution of an old bug, which I think was resolved in USB via https://github.com/SUSE/cf-usb/pull/16

The pre-upgrade usb deploy task doesn't do the same kind of testing on its own, and we've only ever used those two tasks in unison (and haven't even been running those tasks lately). The comments in the flags.yml file give a bit of clarity into the intent:

`# deploy mysql and postgres services, sidecars, and sample apps before an upgrade`

and

`# test apps which use services, and tear down apps, services, and sidecars`

respectively.

The post-upgrade part would *never* be run in a non-upgrade pipeline because the stuff it expected wouldn't exist (and the pre-upgrade task only runs before the deploy task which would run in a non-upgrade pipeline)

We could modify pipeline behaviour to run both tasks on a straight deploy, or even after an upgrade, but since we've deprioritized USB testing, it's not something we've had a card for. As it is, those two tasks were specifically designed to work in unison in upgrade pipelines as an integration test.

I do think we really should have only had one flag which enabled or disabled both of these tasks ( `enable-usb-upgrade-integration` ?)

qa-pipelines/qa-pipeline.yml.erb
HartS: OK, two separate points:

- Scaled HA is not something our pipeline has ever preferred since we switched to using the HA flag. If we had a build which the default scaling of `HA=true` introduced breakages, we would then want to set the scaled HA sizing parameters accordingly and merge a change in our pipeline manifest to use `SCALED_HA` instead of `HA` here. The last time we deployed with the `SCALED_HA` setting was roughly a year ago, when we modified the qa-pipeline file to set `HA: true` instead: https://github.com/SUSE/cf-ci/commit/921e31e30d6a2d11c6d868f32c4c61ad5614d07c . For our template I think we can ignore this, and then in the future if we need to use `SCALED_HA` again for some reason to get more fine-grained control of the scaling, we'd merge the desired parameter changes into the cf-deploy-upgrade-common.sh file accordingly and modify the template to change HA to SCALED_HA

- The pipeline as defined by our manifest deploys two jobs, one SA and one HA. The flags you added allow you to turn one or the other job on or off, but as it is, all jobs are paused by default and we only unpause the one we want to run builds from. Sometimes I'll deploy a pipeline, unpause the SA job and run a build from that, and then later unpause the HA job and run that one too. So I don't think we gain anything from turning these jobs on or off at the template level, and furthermore, it introduces the potential to accidentally deploy a pipeline in which both jobs are disabled (which effectively means the pipeline does nothing)

qa-pipelines/flags.yml
mook-as: Okay, 1ced19a27b16313a31e5263996b5234fd5954a40 added to follow this description (which seems correct).  Called it `enable-cf-usb-upgrade-tests`. Thanks!

qa-pipelines/qa-pipeline.yml.erb
mook-as: Okay, 79d89eb0a6336da9734e06b59fdb31a9a919c816 changed to drop the SA/HA flags. The scaled HA stuff is basically dead code that won't be reached right now.

qa-pipelines/flags.yml
mook-as: README updated in 5c05c5bcf031a6bfe03b100c61301f4f37a274b5.

I think preset files are good (at least, having configs on disk are good) because that makes it easier to reproduce pipelines. But we may want to consider doing different things later… but that's later.

qa-pipelines/qa-pipeline.yml.erb
HartS: Talked about this, it might be better to consolidate the config files into one (which will include the pre-upgrade bundle URL) and have the set-pipeline script add the necessary changes to the pool-specific parameters. set-pipeline can also take a flag to override the *deploy* version for deployments from custom versions (nightlies, PRs, etc.). The pre-upgrade bundle URL was previously being used for this in a hacky way, but the refactor can improve all this

qa-pipelines/generate-vars.rb
```diff
@@ -0,0 +1,68 @@
+#!/usr/bin/env ruby
```
HartS: Based on a reading and some discussion, it looks like this file may not actually be simplifying anything. I think the original intent was to feed something into the generate-pipeline file, but that's not happening now

qa-pipelines/qa-pipeline.yml.erb
```diff
@@ -0,0 +1,253 @@
...
+    <% if enable_cf_acceptance_tests_pre_upgrade %>
+    - task: acceptance-tests-pre-upgrade
+      file: ci/qa-pipelines/tasks/run-test.yml
+      params:
+        << : *upgrade-task-params
+        TEST_NAME: acceptance-tests
+      input_mapping:
+        s3.scf-config: s3.scf-config-sles
+    <% end %> # enable_cf_acceptance_tests_pre_upgrade
+    <% if enable_cf_deploy %>
```
HartS: enable_cf_upgrade

HartS:
Just noticed this text is still the same: https://github.com/SUSE/cf-ci/pull/248/files#diff-28bc02ec21e2a43f8a7705cceb798ccbR24 . Should be `pipeline-presets/single-brain.yml` now

qa-pipelines/flags.yml
```diff
@@ -4,37 +4,39 @@
...
+# (don't deploy), or a string (in which case use given bundle URL)
+enable-cf-deploy: true
 
 # test apps which use services, and tear down apps, services, and sidecars
-enable-usb-post-upgrade: true
-
-# NON-UPGRADE PIPELINES START HERE
-# deploy for non-upgrade pipelines
-enable-cf-deploy: false
+enable-cf-usb: true
```
HartS: I don't believe this is used now

qa-pipelines/flags.yml
```diff
@@ -4,37 +4,39 @@
...
 
 # run tests before an upgrade
 enable-cf-smoke-tests-pre-upgrade: true
 enable-cf-brain-tests-pre-upgrade: true
 
-# This should *only* ever be true when running a *full* pre-upgrade pipeline. The only reason to do this is
-# when you want to deploy a pipeline which uses a custom config URL instead of determining it from the
-# s3.scf-config-* resources. A typical upgrade pipeline only runs cats post-upgrade, and a typical deploy
-# pipeline has all the 'pre-upgrade' flags set to false
+# This should *only* ever be true when running a *full* pre-upgrade pipeline.
```
HartS: With the change of the pre-upgrade deploy to `upgrade-from-version`, this isn't as clear. Actually, I don't think it even works that way because custom bundle deployments are run as normal deployments, but with the custom URL put into `enable-cf-deploy` right?

HartS: In other words we may want to remove `enable-cf-acceptance-tests-pre-upgrade` altogether

qa-pipelines/flags.yml
```diff
@@ -4,37 +4,39 @@
 # The settings in this file will also be valid for our most complete pipeline
 # (which performs an upgrade and all tests)
 
-# deploy cf for upgrade pipelines
-enable-cf-deploy-pre-upgrade: true
+# Report status as steps complete
+status-reporting: true
+
+# version to upgrade from; if null, upgrade is not performed.
+upgrade-from-version: scf-sle-2.15.2+cf3.6.0.0.gde1bd02f
```
HartS: I suspect renaming `enable-cf-deploy-pre-upgrade` to this will lead to accidental enabling on pipelines where that task is not intended to be enabled

qa-pipelines/qa-pipeline.yml.erb
mook-as: But it's `enable-cf-deploy` in [`flags.yml`](https://github.com/SUSE/cf-ci/pull/248/files/5c05c5bcf031a6bfe03b100c61301f4f37a274b5#diff-b3614856767f8ba1e0bf96784e99bc18R31)?

qa-pipelines/generate-vars.rb
mook-as: Reverted in b2c57c912018ff0f716f9ec90bcbb2448a888802 (or rather, will be when I push it); we might want to `git rebase -i` before merge to drop that one as well as the commit introducing the change.

qa-pipelines/flags.yml
mook-as: Fixed in 093a72171aa28cbdc2baefccb26dae7a24abc59c

qa-pipelines/flags.yml
mook-as: Removed in 2682d149dedca87a4db2b3624952b57f9ebf58a0

PR #250, opened 2019-05-27
eks QA CI

prabalsharma:
- Update cf-ci-orchestration image: aws cli install, k install, klog install, aws-iam-authenticator install, update helm cli, update kubectl cli
- add config-eks.yml
- add pipeline-presets/upgrades-NoUsb-NoBrainPreUpgrd.yml 
- add aws jenkins user creds
- add azure login creds
- use aks to create dns record set for eks lbs
- specify tag for splatform/cf-ci-orchestration image

### DIFF
--- a/cf-ci-orchestration/Dockerfile
+++ b/cf-ci-orchestration/Dockerfile
@@ -12,7 +12,6 @@ RUN true \
     && zypper --non-interactive addrepo http://download.opensuse.org/repositories/utilities/openSUSE_Leap_42.3/utilities.repo \
     && zypper --non-interactive --gpg-auto-import-keys refresh \
     && zypper --non-interactive install ruby \
-    python-pip \
     unzip \
     openssh \
     sshpass \
@@ -24,13 +23,24 @@ RUN true \
     direnv \
     bash-completion docker-bash-completion systemd-bash-completion \
     python-xml python-pyOpenSSL python-devel \
-    && pip install awscli \
+    && curl -sS "https://s3.amazonaws.com/aws-cli/awscli-bundle.zip" -o "awscli-bundle.zip" \
+    && unzip awscli-bundle.zip \
+    && ./awscli-bundle/install -i /usr/local/aws -b /usr/local/bin/aws \
+    && rm -rf aws-cli-bundle aws-cli-bundle.zip \
+    && curl -o aws-iam-authenticator https://amazon-eks.s3-us-west-2.amazonaws.com/1.12.7/2019-03-27/bin/linux/amd64/aws-iam-authenticator \
+    && chmod +x ./aws-iam-authenticator \
+    && mv ./aws-iam-authenticator /usr/local/bin/aws-iam-authenticator \
     && rpm --import https://packages.microsoft.com/keys/microsoft.asc \
     && zypper addrepo --name 'Azure CLI' --check https://packages.microsoft.com/yumrepos/azure-cli azure-cli \
     && zypper --non-interactive install --from azure-cli azure-cli \
+    && helm_version=v2.8.2 \
+    && wget https://kubernetes-helm.storage.googleapis.com/helm-${helm_version}-linux-amd64.tar.gz -O - | tar xz -C /usr/local/bin --strip-components=1 linux-amd64/helm \
+    && mv /usr/local/bin/helm /usr/local/bin/helm_${helm_version} \
+    && chmod a+x /usr/local/bin/helm_${helm_version} \
     && helm_version=v2.11.0 \
     && wget https://kubernetes-helm.storage.googleapis.com/helm-${helm_version}-linux-amd64.tar.gz -O - | tar xz -C /usr/local/bin --strip-components=1 linux-amd64/helm \
-    && chmod a+x /usr/local/bin/helm \
+    && mv /usr/local/bin/helm /usr/local/bin/helm_${helm_version} \
+    && chmod a+x /usr/local/bin/helm_${helm_version} \
     && echo "eval \"\$(direnv hook bash)\"" >> ~/.bashrc \
     && echo "source /etc/profile.d/bash_completion.sh" >> ~/.bashrc \
     && echo "source <(helm completion bash)" >> ~/.bashrc \
@@ -38,9 +48,6 @@ RUN true \
     && chmod +x ./kubectl \
     && mv ./kubectl /usr/local/bin/kubectl \
     && echo "source <(kubectl completion bash)" >> ~/.bashrc \
-    && curl -o aws-iam-authenticator https://amazon-eks.s3-us-west-2.amazonaws.com/1.12.7/2019-03-27/bin/linux/amd64/aws-iam-authenticator \
-    && chmod +x ./aws-iam-authenticator \
-    && mv ./aws-iam-authenticator /usr/local/bin/aws-iam-authenticator \
     && curl -sLo /usr/local/bin/klog.sh "https://raw.githubusercontent.com/SUSE/scf/develop/container-host-files/opt/scf/bin/klog.sh" \
     && chmod +x /usr/local/bin/klog.sh \
     && curl -sLo /usr/local/bin/k "https://github.com/aarondl/kctl/releases/download/v0.0.12/kctl-linux-amd64" \
@@ -66,5 +73,9 @@ RUN true \
     && mkdir -p /usr/local/gcloud \
     && tar -C /usr/local/gcloud -xvf /tmp/google-cloud-sdk.tar.gz \
     && /usr/local/gcloud/google-cloud-sdk/install.sh --quiet
+
+COPY helm /usr/local/bin/helm
+RUN true \
+    && chmod a+x /usr/local/bin/helm
 
 ENV PATH $PATH:/usr/local/gcloud/google-cloud-sdk/bin
--- a/cf-ci-orchestration/Dockerfile
+++ b/cf-ci-orchestration/Dockerfile
@@ -12,6 +12,7 @@ RUN true \
     && zypper --non-interactive addrepo http://download.opensuse.org/repositories/utilities/openSUSE_Leap_42.3/utilities.repo \
     && zypper --non-interactive --gpg-auto-import-keys refresh \
     && zypper --non-interactive install ruby \
+    python-pip \
     unzip \
     openssh \
     sshpass \
@@ -23,19 +24,27 @@ RUN true \
     direnv \
     bash-completion docker-bash-completion systemd-bash-completion \
     python-xml python-pyOpenSSL python-devel \
+    && pip install awscli \
     && rpm --import https://packages.microsoft.com/keys/microsoft.asc \
     && zypper addrepo --name 'Azure CLI' --check https://packages.microsoft.com/yumrepos/azure-cli azure-cli \
     && zypper --non-interactive install --from azure-cli azure-cli \
-    && helm_version=v2.8.2 \
+    && helm_version=v2.11.0 \
     && wget https://kubernetes-helm.storage.googleapis.com/helm-${helm_version}-linux-amd64.tar.gz -O - | tar xz -C /usr/local/bin --strip-components=1 linux-amd64/helm \
     && chmod a+x /usr/local/bin/helm \
     && echo "eval \"\$(direnv hook bash)\"" >> ~/.bashrc \
     && echo "source /etc/profile.d/bash_completion.sh" >> ~/.bashrc \
     && echo "source <(helm completion bash)" >> ~/.bashrc \
-    && curl -LO https://storage.googleapis.com/kubernetes-release/release/v1.9.8/bin/linux/amd64/kubectl \
+    && curl -LO https://storage.googleapis.com/kubernetes-release/release/v1.11.8/bin/linux/amd64/kubectl \
     && chmod +x ./kubectl \
     && mv ./kubectl /usr/local/bin/kubectl \
     && echo "source <(kubectl completion bash)" >> ~/.bashrc \
+    && curl -o aws-iam-authenticator https://amazon-eks.s3-us-west-2.amazonaws.com/1.12.7/2019-03-27/bin/linux/amd64/aws-iam-authenticator \
+    && chmod +x ./aws-iam-authenticator \
+    && mv ./aws-iam-authenticator /usr/local/bin/aws-iam-authenticator \
+    && curl -sLo /usr/local/bin/klog.sh "https://raw.githubusercontent.com/SUSE/scf/develop/container-host-files/opt/scf/bin/klog.sh" \
+    && chmod +x /usr/local/bin/klog.sh \
+    && curl -sLo /usr/local/bin/k "https://github.com/aarondl/kctl/releases/download/v0.0.12/kctl-linux-amd64" \
+    && chmod +x /usr/local/bin/k \
     && zypper clean -a \
     && true
 
@@ -56,6 +65,6 @@ RUN true \
     && curl https://dl.google.com/dl/cloudsdk/release/google-cloud-sdk.tar.gz > /tmp/google-cloud-sdk.tar.gz \
     && mkdir -p /usr/local/gcloud \
     && tar -C /usr/local/gcloud -xvf /tmp/google-cloud-sdk.tar.gz \
-    && /usr/local/gcloud/google-cloud-sdk/install.sh --quiet 
+    && /usr/local/gcloud/google-cloud-sdk/install.sh --quiet
 
 ENV PATH $PATH:/usr/local/gcloud/google-cloud-sdk/bin
--- a/qa-pipelines/qa-pipeline.yml
+++ b/qa-pipelines/qa-pipeline.yml
@@ -99,6 +99,11 @@ jobs:
         MAGIC_DNS_SERVICE: ((magic-dns-service))
         CAP_INSTALL_VERSION: ((cap-sle-url))
         ENABLE_CF_DEPLOY_PRE_UPGRADE: ((enable-cf-deploy-pre-upgrade))
+        AZ_SP_APPID: ((az-sp-app-id))
+        AZ_SP_PASSWORD: ((az-sp-password))
+        AZ_SP_TENANT: ((az-sp-tenant))
+        AWS_ACCESS_KEY_ID: ((aws-jenkins-access-key))
+        AWS_SECRET_ACCESS_KEY: ((aws-jenkins-secret-key))
         GKE_PRIVATE_KEY_BASE64: ((gke-private-key-base64))
       input_mapping:
         s3.scf-config: s3.scf-config-sles
@@ -107,6 +112,11 @@ jobs:
       params:
         CAP_INSTALL_VERSION: ((cap-sle-url))
         ENABLE_CF_SMOKE_TESTS_PRE_UPGRADE: ((enable-cf-smoke-tests-pre-upgrade))
+        AZ_SP_APPID: ((az-sp-app-id))
+        AZ_SP_PASSWORD: ((az-sp-password))
+        AZ_SP_TENANT: ((az-sp-tenant))
+        AWS_ACCESS_KEY_ID: ((aws-jenkins-access-key))
+        AWS_SECRET_ACCESS_KEY: ((aws-jenkins-secret-key))
         GKE_PRIVATE_KEY_BASE64: ((gke-private-key-base64))
       input_mapping:
         s3.scf-config: s3.scf-config-sles
@@ -115,6 +125,11 @@ jobs:
       params:
         CAP_INSTALL_VERSION: ((cap-sle-url))
         ENABLE_CF_BRAIN_TESTS_PRE_UPGRADE: ((enable-cf-brain-tests-pre-upgrade))
+        AZ_SP_APPID: ((az-sp-app-id))
+        AZ_SP_PASSWORD: ((az-sp-password))
+        AZ_SP_TENANT: ((az-sp-tenant))
+        AWS_ACCESS_KEY_ID: ((aws-jenkins-access-key))
+        AWS_SECRET_ACCESS_KEY: ((aws-jenkins-secret-key))
         GKE_PRIVATE_KEY_BASE64: ((gke-private-key-base64))
       input_mapping:
         s3.scf-config: s3.scf-config-sles
@@ -123,6 +138,11 @@ jobs:
       params:
         CAP_INSTALL_VERSION: ((cap-sle-url))
         ENABLE_CF_ACCEPTANCE_TESTS_PRE_UPGRADE: ((enable-cf-acceptance-tests-pre-upgrade))
+        AZ_SP_APPID: ((az-sp-app-id))
+        AZ_SP_PASSWORD: ((az-sp-password))
+        AZ_SP_TENANT: ((az-sp-tenant))
+        AWS_ACCESS_KEY_ID: ((aws-jenkins-access-key))
+        AWS_SECRET_ACCESS_KEY: ((aws-jenkins-secret-key))
         GKE_PRIVATE_KEY_BASE64: ((gke-private-key-base64))
       input_mapping:
         s3.scf-config: s3.scf-config-sles
@@ -130,6 +150,11 @@ jobs:
       file: ci/qa-pipelines/tasks/usb-deploy.yml
       params:
         ENABLE_USB_DEPLOY: ((enable-usb-deploy))
+        AZ_SP_APPID: ((az-sp-app-id))
+        AZ_SP_PASSWORD: ((az-sp-password))
+        AZ_SP_TENANT: ((az-sp-tenant))
+        AWS_ACCESS_KEY_ID: ((aws-jenkins-access-key))
+        AWS_SECRET_ACCESS_KEY: ((aws-jenkins-secret-key))
         GKE_PRIVATE_KEY_BASE64: ((gke-private-key-base64))
     - task: cf-upgrade
       file: ci/qa-pipelines/tasks/cf-upgrade.yml
@@ -142,6 +167,11 @@ jobs:
         SCALED_HA: false
         MAGIC_DNS_SERVICE: ((magic-dns-service))
         ENABLE_CF_UPGRADE: ((enable-cf-upgrade))
+        AZ_SP_APPID: ((az-sp-app-id))
+        AZ_SP_PASSWORD: ((az-sp-password))
+        AZ_SP_TENANT: ((az-sp-tenant))
+        AWS_ACCESS_KEY_ID: ((aws-jenkins-access-key))
+        AWS_SECRET_ACCESS_KEY: ((aws-jenkins-secret-key))
         GKE_PRIVATE_KEY_BASE64: ((gke-private-key-base64))
       input_mapping:
         s3.scf-config: s3.scf-config-sles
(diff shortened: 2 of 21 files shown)


cf-ci-orchestration/Dockerfile
```diff
@@ -12,6 +12,7 @@ RUN true \
     && zypper --non-interactive addrepo http://download.opensuse.org/repositories/utilities/openSUSE_Leap_42.3/utilities.repo \
     && zypper --non-interactive --gpg-auto-import-keys refresh \
     && zypper --non-interactive install ruby \
+    python-pip \
```
HartS: Instead of using pip, we might also consider adding the `aws` cli to our orchestration image via the bundled installation so we aren't installing one-off tools for this CLI. Note that this bundle still *does* use pip, but it uses a pip executable within the bundle that also has all necessary dependencies, so is easy to delete.

In a now-deprecated precursor to our CI, we did this:

```
  cd /tmp
  curl -sS "https://s3.amazonaws.com/aws-cli/awscli-bundle.zip" -o "awscli-bundle.zip"
  unzip awscli-bundle.zip
  ./awscli-bundle/install -i /usr/local/aws -b /usr/local/bin/aws
  rm -rf aws-cli-bundle aws-cli-bundle.zip
```

This installs a recent version of the aws CLI for our use

If we go this route, we can delete line 27 as well

cf-ci-orchestration/Dockerfile
```diff
@@ -23,19 +24,27 @@ RUN true \
...
     && chmod a+x /usr/local/bin/helm \
     && echo "eval \"\$(direnv hook bash)\"" >> ~/.bashrc \
     && echo "source /etc/profile.d/bash_completion.sh" >> ~/.bashrc \
     && echo "source <(helm completion bash)" >> ~/.bashrc \
-    && curl -LO https://storage.googleapis.com/kubernetes-release/release/v1.9.8/bin/linux/amd64/kubectl \
+    && curl -LO https://storage.googleapis.com/kubernetes-release/release/v1.11.8/bin/linux/amd64/kubectl \
     && chmod +x ./kubectl \
     && mv ./kubectl /usr/local/bin/kubectl \
     && echo "source <(kubectl completion bash)" >> ~/.bashrc \
+    && curl -o aws-iam-authenticator https://amazon-eks.s3-us-west-2.amazonaws.com/1.12.7/2019-03-27/bin/linux/amd64/aws-iam-authenticator \
```
HartS: It'd be a bit cleaner to move any aws cli installation commands right above this line

qa-pipelines/qa-pipeline.yml
```diff
@@ -99,6 +99,11 @@ jobs:
         MAGIC_DNS_SERVICE: ((magic-dns-service))
         CAP_INSTALL_VERSION: ((cap-sle-url))
         ENABLE_CF_DEPLOY_PRE_UPGRADE: ((enable-cf-deploy-pre-upgrade))
+        AZ_SP_APPID: ((az-sp-app-id))
```
HartS: To limit the number of parameters that need to be passed in, see how we use the GKE private key information.

qa-pipelines/tasks/cf-teardown.sh
```diff
@@ -30,8 +30,12 @@ for namespace in "$CF_NAMESPACE" "$UAA_NAMESPACE" ; do
     done
 done
 
-cap_platform=$(kubectl get configmap -n kube-system cap-values -o json | jq -r .data.platform)
-if [[ ${cap_platform} == "azure" ]] || [[ ${cap_platform} == "gke" ]]; then
+if [[ ${cap_platform} != "eks" ]]; then
```
HartS: I see `cap_platform` now gets set to `eks` in `prepare-kubeconfig.sh` which is sourced, but I think it's unset on non-eks.

Regardless, this isn't correct; we don't want to run `azure_dns_clear` for all deployments, just azure and gke deployments.

qa-pipelines/tasks/lib/azure-aks.sh
```diff
@@ -81,23 +82,43 @@ azure_set_a_record() {
...
+}
+
 azure_set_record_sets_for_namespace() {
     local namespace=$1
+    local cap_platform=$2
     local lb_info lb_ip
-    lb_info=$(kubectl get svc -n $namespace -o json | jq -c '[.items[] | select(.spec.type == "LoadBalancer") | {"svc": .metadata.name, "ip": .status.loadBalancer.ingress[0].ip}]')
+    if [[ ${cap_platform} == "eks" ]]; then
+        lb_info=$(kubectl get svc -n $namespace -o json | jq -c '[.items[] | select(.spec.type == "LoadBalancer") | {"svc": .metadata.name, "ip": .status.loadBalancer.ingress[0].hostname}]')
+        azure_set_record="azure_set_cname_record"
```
HartS: Instead of storing the function name to call in a variable and then indirectly invoking the correct dns function via variable substitution, `azure_set_record` should be a delegation function

qa-pipelines/tasks/lib/azure-aks.sh
```diff
@@ -81,23 +82,43 @@ azure_set_a_record() {
...
+        --cname $2
+}
+
 azure_set_record_sets_for_namespace() {
     local namespace=$1
+    local cap_platform=$2
     local lb_info lb_ip
-    lb_info=$(kubectl get svc -n $namespace -o json | jq -c '[.items[] | select(.spec.type == "LoadBalancer") | {"svc": .metadata.name, "ip": .status.loadBalancer.ingress[0].ip}]')
+    if [[ ${cap_platform} == "eks" ]]; then
+        lb_info=$(kubectl get svc -n $namespace -o json | jq -c '[.items[] | select(.spec.type == "LoadBalancer") | {"svc": .metadata.name, "ip": .status.loadBalancer.ingress[0].hostname}]')
```
HartS: We should use one `lb_info` definition:

```
lb_info=$(
  kubectl get svc -n $namespace -o json | 
  jq -c '
    [
      .items[] | select(.spec.type == "LoadBalancer") | {
        "svc": .metadata.name,
        "ip": .status.loadBalancer.ingress[0].ip,
        "hostname": .status.loadBalancer.ingress[0].hostname
       }
    ]'
)
```

This way you can grab `.hostname` from the `lb_info` element when using EKS, and `.ip` otherwise. For EKS, `.ip` will be null, and for non-EK, `.hostname` will be null

qa-pipelines/tasks/cf-deploy.sh
```diff
@@ -49,11 +51,12 @@ helm install ${CAP_DIRECTORY}/helm/uaa/ \
...
 wait_for_release uaa
 
-if [[ ${cap_platform} == "azure" ]] || [[ ${cap_platform} == "gke" ]]; then
+if [[ ${cap_platform} == "azure" ]] || [[ ${cap_platform} == "gke" ]] ||
+ [[ ${cap_platform} == "eks" ]]; then
     az_login
     azure_dns_clear
-    azure_wait_for_lbs_in_namespace uaa
-    azure_set_record_sets_for_namespace uaa
+    azure_wait_for_lbs_in_namespace uaa ${cap_platform}
```
HartS: I don't think we need to pass in cap_platform as a parameter; the variable is already set in the common lib, which is sourced at the beginning of cf-deploy.sh

qa-pipelines/tasks/lib/azure-aks.sh
```diff
@@ -2,29 +2,24 @@
...
+    if ! az login --service-principal -u ${AZ_SP_APPID} --password ${AZ_SP_PASSWORD} --tenant ${AZ_SP_TENANT}; then
+        echo "Azure login error"
         exit 1
     fi
 }
 
 AZURE_DNS_ZONE_NAME=susecap.net
 AZURE_DNS_RESOURCE_GROUP=susecap-domain
-AZURE_AKS_RESOURCE_GROUP=$(kubectl get configmap -n kube-system -o json cap-values | jq -r '.data["resource-group"]')
+if cat ~/.kube/config | grep "eks.amazonaws.com" > /dev/null ; then
```
HartS: useless use of cat

qa-pipelines/tasks/lib/prepare-kubeconfig.sh
```diff
@@ -5,10 +5,44 @@ read_yaml_key() {
...
+          groups:
+            - system:masters
+EOF
+}
+
 pool_file=${pool_file:-pool.kube-hosts/metadata}
 mkdir -p /root/.kube
 if [[ $(read_yaml_key ${pool_file} kind) == "Config" ]]; then
     cp ${pool_file} /root/.kube/config
+    if cat ~/.kube/config | grep "eks.amazonaws.com" > /dev/null ; then
```
HartS: useless use of cat

qa-pipelines/tasks/lib/prepare-kubeconfig.sh
```diff
@@ -5,10 +5,44 @@ read_yaml_key() {
...
+        cap_platform="eks"
+        if ! cat ~/.kube/config | grep "arn:aws:iam::138384977974:role/eksServiceRole" > /dev/null ||
+         kubectl get configmap -n kube-system aws-auth | grep "arn:aws:iam::138384977974:role/eksServiceRole" > /dev/null; then
+            echo "Your cuslter may not have been configured for eksServiceRole. Make sure you run:"
```
HartS: typo 'cluster'

qa-pipelines/tasks/lib/prepare-kubeconfig.sh
```diff
@@ -5,10 +5,44 @@ read_yaml_key() {
     ruby -r yaml -e "puts YAML.load_file('$1')[\"$2\"]"
 }
 
+aws_auth_cm_yaml() {
```
HartS: Can you add some comments about what this does? And possibly move into a separate file

qa-pipelines/tasks/lib/cf-deploy-upgrade-common.sh
```diff
@@ -165,6 +175,15 @@ set_uaa_sizing_params() {
 }
 
 set_scf_sizing_params() {
+    if [[ ${cap_platform} == "eks" ]] ; then
+        HELM_PARAMS+=(--set sizing.cc_uploader.capabilities={"SYS_RESOURCE"})
```
HartS: All these lines can be combined using a brace expansion

qa-pipelines/tasks/lib/cf-deploy-upgrade-common.sh
```diff
@@ -110,24 +122,25 @@ get_internal_ca_cert() (
...
+    fi
 
     if [[ $(helm_chart_version) == "2.15.2" ]]; then
         HELM_PARAMS+=(--set "sizing.credhub_user.count=1")
     else
         HELM_PARAMS+=(--set "enable.credhub=true")
     fi
 
-    if [[ ${cap_platform} == "azure" ]] || [[ ${cap_platform} == "gke" ]]; then
+    if [[ ${cap_platform} == "azure" ]] || [[ ${cap_platform} == "gke" ]] ||
```
HartS:
```
if [[ ${cap_platform} =~ ^azure$|^gke$|^eks$ ]]; then
...
```

qa-pipelines/tasks/cf-teardown.sh
prabalsharma: we are using azure dns for EKS as well

cf-ci-orchestration/Dockerfile
prabalsharma: I just decided to go with the official way described in Amazon docs

qa-pipelines/tasks/lib/prepare-kubeconfig.sh
prabalsharma: Because EKS clusters will be created by users and then they will have to prepare their kubeconfig for eksServiceRole as in line 38

qa-pipelines/tasks/lib/prepare-kubeconfig.sh
prabalsharma: it is part of comments. From line 37 to 41

qa-pipelines/tasks/lib/cf-deploy-upgrade-common.sh
prabalsharma: yeah, but it gives no advantage. I find this more readable

cf-ci-orchestration/Dockerfile
prabalsharma: why? I didn't get this.

cf-ci-orchestration/Dockerfile
HartS: This is one of several official ways described in their documentation: https://docs.aws.amazon.com/cli/latest/userguide/cli-chap-install.html#install-tool-bundled

cf-ci-orchestration/Dockerfile
HartS: because it makes more sense to put all the commands related to aws tooling in one part of the Dockerfile

qa-pipelines/tasks/cf-teardown.sh
HartS: right, ignore the point about setting the DNS then, I now see that you moved that into a different block just after this one.

What I'm saying is that `cap_platform` won't be set if the platform isn't eks, so this will fail due to `nounset`

Instead this should be:
```
cap_platform=${cap_platform:-$(kubectl get configmap -n kube-system cap-values -o json | jq -r .data.platform)}
```

qa-pipelines/tasks/lib/prepare-kubeconfig.sh
HartS: OK, I'll take another look at this once there are comments explaining why that's needed.

qa-pipelines/tasks/lib/prepare-kubeconfig.sh
HartS: that's not really a comment, and it doesn't explain what eksServiceRole is or why we need it.

qa-pipelines/tasks/lib/cf-deploy-upgrade-common.sh
HartS: Avoiding writing the same thing multiple times is almost always preferable. Same reason we're templating the qa-pipeline.yml file; readability is subjective (and I'd argue the compact form using brace expansion is still more readable), repeating the same thing multiple times introduces more opportunity for error, and makes errors harder to catch. Less unnecessary, repetitive code, and less room for error *is* the advantage.

qa-pipelines/tasks/cf-teardown.sh
HartS: Note that you don't need the `if` clause at all there

cf-ci-orchestration/Dockerfile
prabalsharma: anh, that what you meant by above. I will place this with aws cli

qa-pipelines/qa-pipeline.yml
prabalsharma: I see what you are saying
```
    base64 -d <<< "${GKE_PRIVATE_KEY_BASE64}" > gke-key.json
    export GKE_SERVICE_ACCOUNT_EMAIL=$(jq -r .client_email gke-key.json)
    export GKE_PROJECT_ID=$(jq -r .project_id gke-key.json)
```
The way I see it. its not about how we are doing things in cf-ci repo, it more about how we doing thing in cloudfoundary/secure. All the credential are stored in this way. I guess it because its easier for people to know what we have in there.

HartS: We control the format of the secrets in the gpg-encrypted secrets file used when deploy the pipeline, putting all the credentials needed for a given platform into one key in that secrets file means that we have fewer parameters we have to remember to include

prabalsharma: won't be doing this, as per our discussion

qa-pipelines/tasks/lib/azure-aks.sh
```diff
@@ -2,29 +2,24 @@
...
 set -eu # errexit, nounset
 
-az_login_error_msg() {
-    cat << 'EOF' >& 2
-    kubectl create secret generic -n kube-system aks-dns-sp \
-        --from-literal appId=${AZ_SP_APPID} \
-        --from-literal tenant=${AZ_SP_TENANT} \
-        --from-literal password=${AZ_SP_PASSWORD}
-EOF
+y2j() {
```
HartS: Add this comment to the `y2j` definition

```
# Outputs a json representation of a yml document. For multi-document files, only the first document will be converted
```

qa-pipelines/tasks/lib/prepare-kubeconfig.sh
satadruroy: The IAM user that created the cluster is hardwired in the AWS auth config map. So this is updating the config map to point to a different user/role e.g. eksServiceRole to avoid the non-portability of kubeconfig.

https://docs.aws.amazon.com/eks/latest/userguide/create-kubeconfig.html

cf-ci-orchestration/Dockerfile
```diff
@@ -23,19 +23,35 @@ RUN true \
...
     && chmod a+x /usr/local/bin/helm \
     && echo "eval \"\$(direnv hook bash)\"" >> ~/.bashrc \
     && echo "source /etc/profile.d/bash_completion.sh" >> ~/.bashrc \
     && echo "source <(helm completion bash)" >> ~/.bashrc \
-    && curl -LO https://storage.googleapis.com/kubernetes-release/release/v1.9.8/bin/linux/amd64/kubectl \
+    && curl -LO https://storage.googleapis.com/kubernetes-release/release/v1.11.8/bin/linux/amd64/kubectl \
     && chmod +x ./kubectl \
     && mv ./kubectl /usr/local/bin/kubectl \
     && echo "source <(kubectl completion bash)" >> ~/.bashrc \
+
```
HartS: unnecessary newline, doesn't this cause an error?

prabalsharma: oops, don't know how that happened. It did not cause error. I will fix it though.

qa-pipelines/tasks/lib/azure-aks.sh
```diff
@@ -2,29 +2,25 @@
...
+        echo "Azure login error"
         exit 1
     fi
 }
 
 AZURE_DNS_ZONE_NAME=susecap.net
 AZURE_DNS_RESOURCE_GROUP=susecap-domain
-AZURE_AKS_RESOURCE_GROUP=$(kubectl get configmap -n kube-system -o json cap-values | jq -r '.data["resource-group"]')
+if grep "eks.amazonaws.com" ~/.kube/config; then
+    AZURE_AKS_RESOURCE_GROUP=$(y2j ~/.kube/config clusters | jq -r .clusters[0].name | cut -d / -f 2)
```
HartS: This second param to y2j doesn't get used right?

qa-pipelines/tasks/lib/azure-aks.sh
```diff
@@ -49,16 +45,30 @@ azure_check_lbs_ready_in_namespace() {
...
+        lb_ready_count=$(echo "${lb_info}" | jq -r '[.[] | select(.ip)] | length')
+    fi
     [[ ${lb_count} -eq ${lb_ready_count} ]]
 }
 
 azure_wait_for_lbs_in_namespace() {
     local namespace=$1
     local count=0
-    while ! azure_check_lbs_ready_in_namespace $namespace; do
+    while ! azure_check_lbs_ready_in_namespace $namespace $cap_platform; do
```
HartS: $cap_platform as a parameter isn't being used; the function receiving it gets it because it's set globally, so no need to pass it as a parameter

qa-pipelines/tasks/lib/azure-aks.sh
prabalsharma: anh, good catch!

qa-pipelines/tasks/lib/azure-aks.sh
```diff
@@ -68,45 +78,71 @@ azure_wait_for_lbs_in_namespace() {
...
+            "hostname": .status.loadBalancer.ingress[0].hostname
+           }
+        ]'
+    )
     for lb_svc in $(echo "${lb_info}" | jq -r '.[] | .svc'); do
-        lb_ip=$(echo "${lb_info}" | jq -r '.[] | select(.svc == "'${lb_svc}'").ip')
         if [[ ${lb_svc} == "uaa-uaa-public" ]]; then
-            azure_set_a_record uaa.$AZURE_AKS_RESOURCE_GROUP $lb_ip
-            azure_set_a_record *.uaa.$AZURE_AKS_RESOURCE_GROUP $lb_ip
+            azure_set_record uaa.$AZURE_AKS_RESOURCE_GROUP "${lb_info}" "${lb_svc}"
```
HartS: I don't think `lb_info` needs to be passed to `azure_set_record`. Can't all the information needed be taken from `lb_svc`?

qa-pipelines/tasks/lib/cf-deploy-upgrade-common.sh
```diff
@@ -26,7 +26,19 @@ fi
 # Check that the kube of the cluster is reasonable
 bash ${CAP_DIRECTORY}/kube-ready-state-check.sh kube
 
-PROVISIONER=$(kubectl get storageclasses persistent -o "jsonpath={.provisioner}")
+if kubectl get sc | grep "persistent" > /dev/null ; then
```
HartS: prefer full names (`storageclass` vs `sc`) for readability

HartS: The storageclass name used should always be `persistent`. If it has a different name in EKS, you can create a new sc named persistent with the same definition of whatever sc you want to use.

qa-pipelines/tasks/lib/cf-deploy-upgrade-common.sh
```diff
@@ -26,7 +26,19 @@ fi
...
+    STORAGECLASS="persistent"
+    PROVISIONER=$(kubectl get storageclasses ${STORAGECLASS} -o "jsonpath={.provisioner}")
+    cap_platform=$(kubectl get configmap -n kube-system cap-values -o json | jq -r .data.platform)
+    garden_rootfs_driver=$(kubectl get configmap -n kube-system cap-values -o json | jq -r '.data["garden-rootfs-driver"] // "btrfs"')
+elif [[ ${cap_platform} == "eks" ]] && kubectl get sc | grep gp2 > /dev/null ; then
+    STORAGECLASS="gp2"
```
HartS: use the `gp2` storageclass definition to create `persistent`

qa-pipelines/tasks/lib/azure-aks.sh
prabalsharma: nope. lb_info is needed for ip/hostname

qa-pipelines/tasks/lib/cf-deploy-upgrade-common.sh
```diff
@@ -26,7 +26,19 @@ fi
 # Check that the kube of the cluster is reasonable
 bash ${CAP_DIRECTORY}/kube-ready-state-check.sh kube
 
-PROVISIONER=$(kubectl get storageclasses persistent -o "jsonpath={.provisioner}")
+if kubectl get sc | grep "persistent" > /dev/null ; then
+    STORAGECLASS="persistent"
+    PROVISIONER=$(kubectl get storageclasses ${STORAGECLASS} -o "jsonpath={.provisioner}")
+    cap_platform=$(kubectl get configmap -n kube-system cap-values -o json | jq -r .data.platform)
```
HartS: Set this outside of the if statement:
```
cap_platform=${cap_platform:-$(kubectl get configmap -n kube-system cap-values -o json | jq -r .data.platform)}
```

qa-pipelines/tasks/lib/cf-deploy-upgrade-common.sh
```diff
@@ -51,7 +63,7 @@ wait_for_jobs() {
         # Get the list of all jobs in the helm release, and subtract the value of 'completed' from 'desired'
         # It would be better to parse this from `helm get manifest`, but since that command is broken in helm
         # v2.8.2 (https://github.com/helm/helm/issues/3833) for now we'll parse it from the human-readable status
-        jobs_desired_remaining=$(helm status $release | awk '/==> v1\/Job/ { getline; getline; while (NF>0) { print $2 - $3; getline } }')
+        jobs_desired_remaining=$(helm${helm_ver} status $release | awk '/==> v1\/Job/ { getline; getline; while (NF>0) { print $2 - $3; getline } }')
```
HartS: Can we assume `helm_ver` will *always* be set here?

qa-pipelines/tasks/lib/cf-deploy-upgrade-common.sh
```diff
@@ -110,24 +122,25 @@ get_internal_ca_cert() (
...
-    HELM_PARAMS+=(--set "kube.psp.privileged=suse.cap.psp.privileged")
-}
-
 # Helm parameters common to UAA and SCF, for helm install and upgrades
 set_helm_params() {
     HELM_PARAMS=(--set "env.DOMAIN=${DOMAIN}"
                  --set "secrets.UAA_ADMIN_CLIENT_SECRET=${UAA_ADMIN_CLIENT_SECRET}"
-                 --set "enable.autoscaler=true")
+                 --set "enable.autoscaler=true"
+                 --set "kube.storage_class.persistent=${STORAGECLASS}")
```
HartS: unnecessary if storageclass name is always `persistent`

qa-pipelines/tasks/lib/cf-deploy-upgrade-common.sh
prabalsharma: That can be achieved once we have a deploy-eks script. But right now I have made this script in a way that would let QA use their EKS clusters. Every SUSE doc related to CAP (including Stratos and Metrics) refers to `gp2` as sc.
Also, I don't think we need to follow `storageclass name used should always be persistent`. Its a configurable parameter.

qa-pipelines/tasks/lib/cf-deploy-upgrade-common.sh
prabalsharma: My above comment explains, why not to do this.

qa-pipelines/tasks/lib/cf-deploy-upgrade-common.sh
prabalsharma: no, because there is no configmap named `cap-values ` in EKS clusters

qa-pipelines/tasks/lib/cf-deploy-upgrade-common.sh
prabalsharma: yes. its set in `/lib/prepare-kubeconfig.sh`

qa-pipelines/tasks/lib/cf-deploy-upgrade-common.sh
prabalsharma: not ture.

qa-pipelines/tasks/lib/cf-deploy-upgrade-common.sh
prabalsharma: EKS CI needs to support all EKS cluster created by referring just SUSE docs. 
Once we have `deploy-eks` script/terraform, we can then take a position that QA CI only works with clusters created by our script. Then we can totally assume that SC is going to be named `persistent`, but until them we will have to make CI compatible with sc name `gp2` as this is what our docs say.

qa-pipelines/tasks/lib/azure-aks.sh
HartS: Right, I see that you're taking just the `.svc` value from the load balancer service structure. This is unintuitive because you're passing an entire array (`lb_info`) and then one piece of information from the element of the array we care (the service information). Instead, we should just pass the one element from the array. `azure_set_record` *only* needs information about the service it's setting a record for, not all the other service information.

Here's what you can do for the above `for` loop instead (relying on the fact that the compact representation won't have any whitespace that isn't in a string value, and `.svc`, `.ip`, and `.hostname` will never have whitespace either):

```
for lb_svc in $(echo "${lb_info}" | jq -c .[]; do
...
  azure_set_record ... "${lb_svc}"
...
done
```

`azure_set_record` can then start with:
```
lb_name=$(jq .svc <<< $2)
lb_hostname=$(jq .hostname <<< $2)
lb_ip=$(jq .ip <<< $2)
```

qa-pipelines/tasks/lib/cf-deploy-upgrade-common.sh
HartS: resolved now with changes to helm cli invocation.

HartS:
All changes now resolved. Merging

PR #253, opened 2019-06-05
Update aks kube version

HartS:
This bypasses issues with installation of the azure open-service-broker
reported in https://github.com/SUSE/doc-cap/issues/448


prabalsharma:
will it be compatible with kubectl client v1.11.8 ?

PR #256, opened 2019-06-06
fix bundle_url

prabalsharma:
- replace ((s3-config-prefix-sles)) to `master` in  bundle_url as otherwise it uses nightly bucket for master build when --nightly flag is used in upgrades CI
- update upgrade version from 2.15.2 to 2.16.4
- remove cap-sle-url from config-{pool}.yml(s)

### DIFF
--- a/qa-pipelines/qa-pipeline.yml.erb
+++ b/qa-pipelines/qa-pipeline.yml.erb
@@ -7,7 +7,7 @@ def bundle_url_from_version(version)
     when true    then '""'      # Use bucket version
     when false   then '""'      # Don't install (non-upgrade pipelines, for example)
     when /:\/\// then version # URL
-    else "https://((s3-config-bucket-sles)).s3.amazonaws.com/((s3-config-prefix-sles))#{ERB::Util.url_encode(version)}.zip"
+    else "https://((s3-config-bucket-sles)).s3.amazonaws.com/#{(version)}.zip"
   end
 end
 
--- a/qa-pipelines/config-azure.yml
+++ b/qa-pipelines/config-azure.yml
@@ -23,4 +23,3 @@ organization: splatform
 magic-dns-service: omg.howdoi.website
 
 src-repo: SUSE/scf
-cap-sle-url: https://s3.amazonaws.com/cap-release-archives/master/scf-sle-2.16.4%2Bcf6.10.0.2.g5abdb16f.zip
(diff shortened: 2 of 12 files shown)


qa-pipelines/qa-pipeline.yml.erb
```diff
@@ -7,7 +7,7 @@ def bundle_url_from_version(version)
     when true    then '""'      # Use bucket version
     when false   then '""'      # Don't install (non-upgrade pipelines, for example)
     when /:\/\// then version # URL
-    else "https://((s3-config-bucket-sles)).s3.amazonaws.com/((s3-config-prefix-sles))#{ERB::Util.url_encode(version)}.zip"
+    else "https://s3.amazonaws.com/((s3-config-bucket-sles))/master/#{ERB::Util.url_encode(version)}.zip"
```
mook-as: AWS has deprecated the `s3.amazonaws.com` version: https://aws.amazon.com/blogs/aws/amazon-s3-path-deprecation-plan-the-rest-of-the-story/

You're changing common code, which means `s3-config-prefix-sles` will basically never be used when deploying non-upgrade pipelines at a specific version.  We need a different fix here (to pass down the correct prefix, possibly as part of the version).

prabalsharma: Ok. I have created card for this https://jira.suse.com/browse/CAP-365

prabalsharma:
> You're changing common code, which means `s3-config-prefix-sles` will basically never be used when deploying non-upgrade pipelines at a specific version. We need a different fix here (to pass down the correct prefix, possibly as part of the version).

@mook-as This is what I understand: CAP bundle url is used only for upgrade CI. 
If my understanding is correct, then we can hard code `master` in url in a way I did. As we never test an upgrade from anything but last released version ( i.e. master/cap_version). Scenario in which we test a specific build version is only `CAP Deploy` and for that we use `--pr` which won't use this `bundle_url` at all.

mook-as:
@prabalsharma No, the pipeline reorg PR made it so you can use the bundle URL for non-upgrade pipelines too (so you don't need to do the weird hack where you "upgrade" from a bundle, and then not actually do the upgrade; that made no sense).

prabalsharma:
As a user can use any directory in a bucket. One will have to specify that in enable-cf-deploy-pre-upgrade along with exact build

PR #259, opened 2019-06-10
eks kpatch

prabalsharma:
CF TCP Route Service opens its ports only on request by CF itself. This does not work with AWS as if the first declared port is not open, then the LB fails the health checks done by AWS itself and is reported as offline. To workaround this, we need to apply this kubectl patch

https://jira.suse.com/browse/CAP-247

### DIFF
--- a/qa-pipelines/tasks/lib/cf-deploy-upgrade-common.sh
+++ b/qa-pipelines/tasks/lib/cf-deploy-upgrade-common.sh
@@ -81,14 +81,14 @@ wait_for_jobs() {
     for job in ${jobs_in_namespace}; do
         echo "waiting for job ${job}"
         seconds_remaining=$(( 4800 + ${start} - $(date +%s) ))
-        set +o errexit 
+        set +o errexit
         kubectl wait job ${job} --namespace ${namespace} --for=condition=complete --timeout ${seconds_remaining}s
         kubectl_wait_status=$?
-        set -o errexit 
+        set -o errexit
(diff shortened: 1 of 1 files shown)


qa-pipelines/tasks/lib/cf-deploy-upgrade-common.sh
```diff
@@ -118,6 +118,12 @@ wait_for_release() {
         printf "%s pods are still pending after 80 minutes \n" "${release}"
         exit 1
     fi
+    if [[ ${cap_platform} == "eks" &&  ${namespace} == "scf" ]]; then
+        healthcheck_port=$(kubectl get svc tcp-router-tcp-router-public '-o=jsonpath={.spec.ports[?(@.name == "healthcheck")].port}' --namespace scf)
```
HartS: `kubectl get service`.. especially because you're using the non-abbreviated form on line 124

HartS: Also I would do `-o='...'` rather than `'-o=...'`

HartS:
Only one small nit. Can you also amend the commit to include some explanation about what it does and why? `eks kpatch` is not an explanatory commit message. The first line of the commit message should ideally be 50 characters or less explaining what the commit does in present tense (e.g. `sets a port for the tcp service if there is non already'). After a blank line, the remainder of the commit message should explain *why* it was doing that (where was this problem occurring and what does it fix)

PR #269, opened 2019-07-23
concourse terraform

prabalsharma:
Integrates terraform to existing qa pipeline
Updates upgrade url to 1.4.1

### DIFF
--- a/cap-terraform/.gitignore
+++ b/cap-terraform/.gitignore
@@ -1,0 +1,9 @@
+**/creds/*
+.terraform
+*.backup
+*.tfstate
+scf*/
+*.info
+**/*cluster-plan/*
+*.tfvars
+temp/
--- a/cap-terraform/.gitignore
+++ b/cap-terraform/.gitignore
@@ -5,5 +5,4 @@
 scf*/
 *.info
 **/*cluster-plan/*
-*.tfvars
 temp/
--- a/cf-ci-orchestration/Dockerfile
+++ b/cf-ci-orchestration/Dockerfile
@@ -2,7 +2,7 @@
 FROM opensuse/leap:15.1
 
 # Install: ruby, unzip, openssh, sshpass, vim, nano, curl, kubectl
-# Install: wget, tar, helm, direnv
+# Install: wget, tar, helm, direnv, terraform, aws-cli, aws-iam-authenticator
 # Install python dependencies for az cli prerequisites
 # "zypper removerepo "OSS Update" is now required because that repo isn't reachable
 # NOTE: The `utilities` repo is added for access to `direnv`.
@@ -24,6 +24,10 @@ RUN true \
     git-core \
     bash-completion docker-bash-completion systemd-bash-completion \
     python-xml python-pyOpenSSL python-devel \
+    && curl -LO "https://releases.hashicorp.com/terraform/0.12.2/terraform_0.12.2_linux_amd64.zip" \
+    && unzip terraform_0.12.2_linux_amd64.zip \
+    && mv ./terraform /usr/local/bin/terraform \
+    && rm terraform_0.12.2_linux_amd64.zip \
     && curl -sS "https://s3.amazonaws.com/aws-cli/awscli-bundle.zip" -o "awscli-bundle.zip" \
     && unzip awscli-bundle.zip \
     && ./awscli-bundle/install -i /usr/local/aws -b /usr/local/bin/aws \
--- a/qa-pipelines/flags.yml
+++ b/qa-pipelines/flags.yml
@@ -11,7 +11,7 @@ enable-terraform: true
 status-reporting: true
 
 # version to upgrade from; if null, upgrade is not performed.
-enable-cf-deploy-pre-upgrade: master/scf-sle-2.16.4%2Bcf6.10.0.2.g5abdb16f
+enable-cf-deploy-pre-upgrade: master/scf-sle-2.17.1%2Bcf7.11.0.4.ga6c993eb
 
 # run tests before an upgrade
 enable-cf-smoke-tests-pre-upgrade: true
--- a/qa-pipelines/flags.yml
+++ b/qa-pipelines/flags.yml
@@ -4,7 +4,7 @@
 # The settings in this file will also be valid for our most complete pipeline
 # (which performs an upgrade and all tests)
 
-# Use terrafrom to deploy AKS cluster
+# Use terraform to deploy AKS cluster
 enable-terraform: true
 
 # Report status as steps complete
--- a/qa-pipelines/flags.yml
+++ b/qa-pipelines/flags.yml
@@ -4,6 +4,9 @@
 # The settings in this file will also be valid for our most complete pipeline
 # (which performs an upgrade and all tests)
 
+# Use terrafrom to deploy AKS cluster
+enable-terraform: true
+
 # Report status as steps complete
 status-reporting: true
 
--- a/qa-pipelines/qa-pipeline.yml.erb
+++ b/qa-pipelines/qa-pipeline.yml.erb
@@ -152,8 +152,8 @@ jobs:
         state: pending
     <% end %> # status_reporting
     on_failure:
+      <% if enable_terraform %>
       try:
-        <% if enable_terraform %>
         do:
           - task: delete-aks-resource-group
             file: ci/qa-pipelines/tasks/delete-aks-resource-group.yml
@@ -161,7 +161,7 @@ jobs:
               << : *common-task-params
           - put: pool.kube-hosts
             params: {remove: pool.kube-hosts}
-        <% end %> # enable_terraform
+      <% end %> # enable_terraform
       <% if !enable_terraform %>
       put: pool.kube-hosts
       params: {release: pool.kube-hosts}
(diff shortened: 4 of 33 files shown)


qa-pipelines/qa-pipeline.yml.erb
```diff
@@ -227,26 +255,38 @@ jobs:
...
     # block, so that when tests fail we have a chance of examining why things are
     # failing.
-    <% if enable_cf_teardown %>
+    <% if enable_cf_teardown && !enable_terraform %>
     - task: cf-teardown
       file: ci/qa-pipelines/tasks/cf-teardown.yml
       timeout: 1h
       params: *common-task-params
-    <% end %> # enable_cf_teardown
+    <% end %> # enable_cf_teardown and not enable_terraform
```
HartS: `and` and `&&` are not operationally equivalent in ruby, so let's stick with `&&` to reflect the condition being closed

cap-terraform/.gitignore
```diff
@@ -0,0 +1,8 @@
+**/creds/*
+.terraform
+*.backup
+*.tfstate
+scf*/
```
HartS: Where are files matching this path prefix created?

cf-ci-orchestration/Dockerfile
```diff
@@ -24,6 +24,10 @@ RUN true \
     git-core \
     bash-completion docker-bash-completion systemd-bash-completion \
     python-xml python-pyOpenSSL python-devel \
+    && curl -LO "https://releases.hashicorp.com/terraform/0.12.2/terraform_0.12.2_linux_amd64.zip" \
```
HartS: When possible, changes to the Dockerfile should happen at the end in a new run step to leverage caching, and avoid rebuilding layers (faster pushes to docker hub and faster pulls from docker hub)

qa-pipelines/flags.yml
```diff
@@ -4,11 +4,14 @@
 # The settings in this file will also be valid for our most complete pipeline
 # (which performs an upgrade and all tests)
 
+# Use terrafrom to deploy AKS cluster
```
HartS: terrafrom

HartS: terrafrom

qa-pipelines/qa-pipeline.yml.erb
```diff
@@ -91,6 +91,15 @@ aliases:
     KUBE_ORGANIZATION: ((organization))
     MAGIC_DNS_SERVICE: ((magic-dns-service))
 
+    # Terraform AKS Configuration
+    ARM_CLIENT_ID: ((arm-client-id))
```
HartS: If these vars are only needed by the terraform task, they shouldn't go in the common task params

qa-pipelines/qa-pipeline.yml.erb
```diff
@@ -135,8 +152,20 @@ jobs:
...
+      <% if enable_terraform %>
+      try:
+        do:
+          - task: delete-aks-resource-group
+            file: ci/qa-pipelines/tasks/delete-aks-resource-group.yml
+            params:
+              << : *common-task-params
+          - put: pool.kube-hosts
+            params: {remove: pool.kube-hosts}
+      <% end %> # enable_terraform
```
HartS: Use an else here instead

qa-pipelines/qa-pipeline.yml.erb
```diff
@@ -135,8 +152,20 @@ jobs:
...
+      <% if !enable_terraform %>
       put: pool.kube-hosts
       params: {release: pool.kube-hosts}
+      <% end %> # not enable_terraform
```
HartS: `# enable_terraform`, not `# not enable_terraform`

qa-pipelines/qa-pipeline.yml.erb
```diff
@@ -227,26 +255,38 @@ jobs:
...
     - task: cf-teardown
       file: ci/qa-pipelines/tasks/cf-teardown.yml
       timeout: 1h
       params: *common-task-params
-    <% end %> # enable_cf_teardown
+    <% end %> # enable_cf_teardown and not enable_terraform
+    <% if !enable_terraform %>
     - put: pool.kube-hosts
       params: {release: pool.kube-hosts}
+    <% end %> # not enable_terraform
```
HartS: Same as above, use `# !enable_terraform`

qa-pipelines/qa-pipeline.yml.erb
```diff
@@ -227,26 +255,38 @@ jobs:
...
     <% end %> # status_reporting
     <% if status_reporting %>
     on_failure:
       put: status.src
       params:
-        <<: *status-params-<%= avail %>
+        << : *status-params-<%= avail %>
         state: failure
     <% end %> # status_reporting
+    <% if enable_terraform %>
```
HartS: Shouldn't terraform teardown task templating be informed by the `enable-teardown` flag also?

qa-pipelines/tasks/delete-aks-resource-group.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
```
HartS: Why can't we just use terraform for teardown?

qa-pipelines/tasks/delete-aks-resource-group.yml
```diff
@@ -0,0 +1,22 @@
...
+inputs:
+- name: ci
+- name: pool.kube-hosts
+
+params:
+  KUBE_REGISTRY_HOSTNAME: ""
+  KUBE_REGISTRY_USERNAME: ""
+  KUBE_REGISTRY_PASSWORD: ""
+  KUBE_ORGANIZATION: ""
+  HA: false # true, false, or "scaled"
```
HartS: annotations aren't needed here. Also, is this param even being used by anything in the delete-terraform task?

cap-terraform/.gitignore
prabalsharma: thats from parent directory: https://github.com/SUSE/cap-terraform/blob/master/.gitignore
I won't touch that as Satadru experimented with that. If that gets cleaned in cap-terraform I will remove it from here

cf-ci-orchestration/Dockerfile
prabalsharma: Do you want a change on this?

qa-pipelines/flags.yml
prabalsharma: oops

prabalsharma: lol
thanks

qa-pipelines/qa-pipeline.yml.erb
prabalsharma: I think we should keep just one prams for all tasks. It keeps the qa-pipeline.erb clean. I was planning to put every thing in common params after refactor2 merge.

qa-pipelines/qa-pipeline.yml.erb
prabalsharma: no its  <% if !enable_terraform %>

qa-pipelines/qa-pipeline.yml.erb
prabalsharma: I kept it this way : 

1.  we can just add additional condition in addition to !enable_terraform when required.
2. keep it consistent with linehttps://github.com/SUSE/cf-ci/blob/d78b41d1862cb40665707215fc3bf0d4cbc30145/qa-pipelines/qa-pipeline.yml.erb#L165
3. easy to read compared to else. At this point, I prefer readability in qa-pipeline.erb we desperately need it :)

qa-pipelines/qa-pipeline.yml.erb
prabalsharma: anh u meant # !enable_terraform

qa-pipelines/qa-pipeline.yml.erb
prabalsharma: no. The terraform CI work follow this guideline:  "QA CI don't care if user want to enable-teardown or not. CI will tear it down."

qa-pipelines/tasks/delete-aks-resource-group.sh
prabalsharma: Cause that will require us to use public cloud to save the state of terraform. Which we don't want to do. An architectural decision was made to not use such resource as its an over kill for our needs. 
In addition, terrform  teardown might leave leftover resources as seen in past whereas use az to delete resource group is very effective.

qa-pipelines/tasks/delete-aks-resource-group.yml
prabalsharma: I just copied this from cf-teardown.yml. As the requirement of such annotations were triggered when using new terrform which I am not using.
Please let me know which line to delete here.

cf-ci-orchestration/Dockerfile
HartS: No it's fine, let's just keep it in mindfor future changes here

qa-pipelines/qa-pipeline.yml.erb
HartS: This is only needed for one terraform task, so it should go in that task parameter list. It's not a common param, and it puts unnecessary things in the pipeline definition (after templating).

qa-pipelines/flags.yml
HartS: Change requested

HartS: change requested

qa-pipelines/qa-pipeline.yml.erb
HartS: Change requested

qa-pipelines/qa-pipeline.yml.erb
HartS: Change requested

qa-pipelines/qa-pipeline.yml.erb
HartS: change requested

PR #272, opened 2019-07-31
qa-pipelines: Try to make waiting for releases more reliable

mook-as:
Try to ensure that we have the secrets in place before continuing.

Or at least be more explicit about when we fail to get the secret.

jsc#CAP-480

### DIFF
--- a/qa-pipelines/tasks/lib/cf-deploy-upgrade-common.sh
+++ b/qa-pipelines/tasks/lib/cf-deploy-upgrade-common.sh
@@ -56,9 +56,8 @@ is_namespace_ready() {
 
     # Check that all pods not from jobs are ready
     if ! kubectl get pods --namespace "${namespace}" --selector '!job-name' \
-        --output 'custom-columns=:.metadata.name,:.status.containerStatuses[].name,:.status.containerStatuses[].ready' \
-        | awk '{ if ($NF != "true") { print } }' \
-        | tee /dev/stderr \
+        --output 'custom-columns=:.status.containerStatuses[*].ready' \
+        | grep false \
         | xargs --no-run-if-empty false
     then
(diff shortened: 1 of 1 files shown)


prabalsharma:
@mook-as  release conflicts

mook-as:
Rebased in 7954917c39ba28454b3726a4853f61135a1402f7.

prabalsharma:
LGTM. @HartS will take a look at it before merging this

qa-pipelines/tasks/lib/cf-deploy-upgrade-common.sh
```diff
@@ -54,13 +54,17 @@ UAA_ADMIN_CLIENT_SECRET="$(head -c32 /dev/urandom | base64)"
...
-        | grep -vE 'secret-generation|post-deployment' \
-        | awk '{ print $2 }' \
-        | sed '/^ *$/d' \
-        | sort \
-        | uniq) ]]
+    # Check that all pods not from jobs are ready
+    if ! kubectl get pods --namespace "${namespace}" --selector '!job-name' \
+        --output 'custom-columns=:.metadata.name,:.status.containerStatuses[].name,:.status.containerStatuses[].ready' \
+        | awk '{ if ($NF != "true") { print } }' \
+        | tee /dev/stderr \
```
mook-as: Just to remind me why this is stuck:
@HartS wants to drop the `tee` here, and instead manually print out what is wrong at the caller, so that we do not fill up logs with errors.

mook-as: Fixed in 86685d420c33614583d39909b44ca8577c2660b9 (plus actually look at _all_ containers).

PR #274, opened 2019-07-31
[cap 370] Modified CI pipeline to set larger timeouts for CC/Broker and Brain tests

andreas-kupries:
Ref: https://jira.suse.com/browse/CAP-370

Minibroker may need 5 minutes to assemble/deliver its catalog, in slow networks
  * CC must not time out, now given 10 minutes to get the catalog.
  * Brain tests have to be allowed to run longer due this also, now have 20 minutes/test.
    I.e. even when broker goes to the border in delivering the catalog the
    remainder of the test will have its regular 10 minutes to complete.


HartS:
LGTM

PR #275, opened 2019-08-01
Changes for PXC migration tests

bikramnehra:
TODO


qa-pipelines/tasks/lib/cf-deploy-upgrade-common.sh
```diff
@@ -214,6 +214,24 @@ set_uaa_sizing_params() {
...
+            HELM_PARAMS+=(--set=sizing.mysql.count=1)
+            return 0    
+        fi
+        return 1
+    fi
+}
+
+pxc_post_upgrade() {
+    if [[ "${HA}" == true ]]; then
+        HELM_PARAMS+=(--set=sizing.mysql.count=3)
```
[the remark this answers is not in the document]
bikramnehra: @jandubois could you provide your feedback on this?

qa-pipelines/tasks/cf-upgrade.sh
```diff
@@ -87,6 +87,39 @@ helm upgrade scf ${CAP_DIRECTORY}/helm/cf/ \
 # Wait for CF release
 wait_for_release scf
 
+if pxc_post_upgrade; then
+  # Delete left-over PVCs from mysql upgrade.
+  kubectl delete pvc -n scf mysql-data-mysql-1
```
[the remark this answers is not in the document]
bikramnehra: HA count for mysql in 1.4.1 was 2, with PXC changes the count will be bumped to 3:

https://github.com/SUSE/scf/pull/2622/files#diff-a65976ef35c13467fe140d46c6ca0d3aL429

qa-pipelines/tasks/lib/cf-deploy-upgrade-common.sh
```diff
@@ -214,6 +214,24 @@ set_uaa_sizing_params() {
     fi
 }
 
+pxc_pre_upgrade() {
```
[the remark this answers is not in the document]
bikramnehra: added condition.

qa-pipelines/tasks/lib/cf-deploy-upgrade-common.sh
```diff
@@ -214,6 +214,24 @@ set_uaa_sizing_params() {
...
+    if [[ "${HA}" == true ]]; then
+        if semver_is_gte 2.17.1 "$(helm_chart_version)"; then
+            HELM_PARAMS+=(--set=sizing.mysql.count=1)
+            return 0    
+        fi
+        return 1
+    fi
+}
+
+pxc_post_upgrade() {
```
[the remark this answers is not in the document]
bikramnehra: added condition.

qa-pipelines/tasks/lib/cf-deploy-upgrade-common.sh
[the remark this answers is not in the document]
jandubois: Yes, this should work. You would only need `--set=config.HA_strict=false` if the minimum number of `mysql` pods for HA mode would be larger than `3`, which seems rather unlikely.

The *only* purpose of `HA_strict` is to disable the minimum instance check, currently:

```
{{- if and .Values.config.HA .Values.config.HA_strict (ne (typeOf .Values.sizing.mysql.count) "<nil>") (lt (int .Values.sizing.mysql.count) 2) }}
_minHAReplicas: {{ fail "mysql must have at least 2 instances for HA" }}
{{- end }}
```

So right now you would only need it if you wanted to scale down `mysql` to a single instance while still setting `config.HA=true`.

qa-pipelines/tasks/cf-deploy.sh
```diff
@@ -52,13 +52,25 @@ if [[ ${cap_platform} =~ ^azure$|^gke$|^eks$ ]]; then
     azure_set_record_sets_for_namespace uaa
 fi
 
+if pxc_pre_upgrade; then
+    echo "Downsizing UAA mysql node count to 1..."
+    echo "${HELM_PARAMS[@]}" | sed 's/kube\.registry\.password=[^[:space:]]*/kube.registry.password=<REDACTED>/g'
+    helm upgrade uaa ${CAP_DIRECTORY}/helm/uaa/ \
```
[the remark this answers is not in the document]
bikramnehra: After discussion with @prabalsharma removed the `return 0` and enabling HA for UAA.

```
set_uaa_sizing_params() {
    return 0 # Sizing UAA up breaks CATS
```

qa-pipelines/tasks/cf-upgrade.sh
```diff
@@ -92,6 +92,47 @@ helm upgrade scf ${CAP_DIRECTORY}/helm/cf/ \
...
+    fi
+    return 1
+}
+
+if pxc_post_upgrade; then
+  # Delete left-over PVCs from mysql upgrade.
+  kubectl delete pvc -n scf mysql-data-mysql-1
+
+  # Scaling the instances up after mysql to pxc migration.
+  echo "Scaling up the instances up for UAA..."
```
HartS: minor typo

qa-pipelines/tasks/cf-deploy.sh
```diff
@@ -94,4 +116,23 @@ if [[ ${cap_platform} =~ ^azure$|^gke$|^eks$ ]]; then
...
+if pxc_pre_upgrade; then
+    echo "Downsizing SCF mysql node count to 1..."
+    echo "${HELM_PARAMS[@]}" | sed 's/kube\.registry\.password=[^[:space:]]*/kube.registry.password=<REDACTED>/g'
+    helm upgrade scf ${CAP_DIRECTORY}/helm/cf/ \
+        --namespace "${CF_NAMESPACE}" \
+        --timeout 3600 \
+        --set "secrets.CLUSTER_ADMIN_PASSWORD=${CLUSTER_ADMIN_PASSWORD:-changeme}" \
+        --set "env.UAA_HOST=${UAA_HOST}" \
+        --set "env.UAA_PORT=${UAA_PORT}" \
+        --set "secrets.UAA_CA_CERT=${CA_CERT}" \
```
HartS: HELM_PARAMS contains the correct value for this, but `CA_CERT` isn't set and causes an error

qa-pipelines/tasks/cf-upgrade.sh
```diff
@@ -92,6 +92,47 @@ helm upgrade scf ${CAP_DIRECTORY}/helm/cf/ \
...
+  
+  echo "Scaling up the instances up for SCF..."
+  echo "${HELM_PARAMS[@]}" | sed 's/kube\.registry\.password=[^[:space:]]*/kube.registry.password=<REDACTED>/g'
+  helm upgrade scf ${CAP_DIRECTORY}/helm/cf/ \
+      --namespace "${CF_NAMESPACE}" \
+      --timeout 3600 \
+      --set "secrets.CLUSTER_ADMIN_PASSWORD=${CLUSTER_ADMIN_PASSWORD:-changeme}" \
+      --set "env.UAA_HOST=${UAA_HOST}" \
+      --set "env.UAA_PORT=${UAA_PORT}" \
+      --set "secrets.UAA_CA_CERT=${CA_CERT}" \
```
HartS: Same note about CA_CERT as above

HartS:
Doesn't `config.HA_strict` need to be disabled for this to work post-upgrade? Since the mysql count is 1

HartS:
Also, I wonder if we should either account for the possibility of embedded UAA or disable embedded UAA in upgrade pipelines until after the workaround is no longer necessary.

qa-pipelines/tasks/cf-upgrade.sh
```diff
@@ -48,10 +48,26 @@ instance_count=$(kubectl get statefulsets -o json diego-cell --namespace scf | j
 monitor_file=$(mktemp -d)/downtime.log
 monitor_url "http://go-env.${DOMAIN}" "${monitor_file}" &
 
+pxc_post_upgrade() {
```
HartS: Just `[[ "${HA}" == true ]]` in this function would be sufficient, and also clearer that this function is just an alias for this check (as opposed to the inversion or something)

Also a nit: the indentation here doesn't match the rest of the file

qa-pipelines/tasks/cf-upgrade.sh
```diff
@@ -70,6 +86,14 @@ fi
 set_helm_params # Resets HELM_PARAMS
 set_scf_sizing_params # Adds scf sizing params to HELM_PARAMS
 
+if pxc_post_upgrade; then
+  #HELM_PARAMS+=(--set=config.HA_strict=false)
+  HELM_PARAMS+=(--set=sizing.mysql.count=1)
+fi
+
+echo SCF customization ...
```
HartS: This is further down now.. delete these two lines

qa-pipelines/tasks/lib/cf-deploy-upgrade-common.sh
```diff
@@ -229,37 +229,29 @@ set_helm_params() {
...
         fi
-    elif [[ ${SCALED_HA} == true ]]; then
-        HELM_PARAMS+=(--set=sizing.{uaa,mysql,mysql_proxy}.count=3)
     fi
 }
 
 set_scf_sizing_params() {
     if [[ ${cap_platform} == "eks" ]] ; then
         HELM_PARAMS+=(--set=sizing.{cc_uploader,nats,routing_api,router,diego_brain,diego_api,diego_ssh}.capabilities[0]="SYS_RESOURCE")
-    fi
```
HartS: This needs to be here still. It's independent of whether `"${HA}" == true`

qa-pipelines/tasks/cf-deploy.sh
```diff
@@ -100,4 +113,28 @@ if [[ ${cap_platform} =~ ^azure$|^gke$|^eks$ ]]; then
     azure_set_record_sets_for_namespace scf
 fi
 
+set_helm_params # Resets HELM_PARAMS
+
+if pxc_pre_upgrade; then
+    echo "Downsizing UAA mysql node count to 1..."
+    echo "${HELM_PARAMS[@]}" | sed 's/kube\.registry\.password=[^[:space:]]*/kube.registry.password=<REDACTED>/g'
```
HartS: printing HELM_PARAMS here is misleading because we're not even using them

qa-pipelines/tasks/cf-deploy.sh
```diff
@@ -100,4 +113,28 @@ if [[ ${cap_platform} =~ ^azure$|^gke$|^eks$ ]]; then
...
+    helm upgrade --reuse-values uaa ${CAP_DIRECTORY}/helm/uaa/ \
+        --namespace "${UAA_NAMESPACE}" \
+        --timeout 600 \
+        --set "sizing.mysql.count=1"
+
+    # Wait for UAA release
+    wait_for_release uaa
+
+    echo "Downsizing SCF mysql node count to 1..."
+    echo "${HELM_PARAMS[@]}" | sed 's/kube\.registry\.password=[^[:space:]]*/kube.registry.password=<REDACTED>/g'
```
HartS: same as above comment

qa-pipelines/tasks/cf-deploy.sh
```diff
@@ -100,4 +113,28 @@ if [[ ${cap_platform} =~ ^azure$|^gke$|^eks$ ]]; then
     azure_set_record_sets_for_namespace scf
 fi
 
+set_helm_params # Resets HELM_PARAMS
```
HartS: I don't believe this is needed actually... we're not using helm params any more

qa-pipelines/tasks/lib/cf-deploy-upgrade-common.sh
```diff
@@ -229,37 +229,29 @@ set_helm_params() {
...
 
 set_uaa_sizing_params() {
-    return 0 # Sizing UAA up breaks CATS
     if [[ "${HA}" == true ]]; then
-        if semver_is_gte "$(helm_chart_version)" 2.11.0; then
+        if [[ ${SCALED_HA} == true ]]; then
+            HELM_PARAMS+=(--set=sizing.{uaa,mysql}.count=2)
+        else
             # HA UAA not supported prior to 2.11.0
             HELM_PARAMS+=(--set=config.HA=true)
```
HartS: We don't want this at all for UAA right now, right?

HartS: OK, based on discussion, we need separate flags for custom UAA sizing and custom SCF sizing (with better names than `SCALED_HA`).

Proposed `CUSTOM_UAA_SIZING` and `CUSTOM_SCF_SIZING`

Because, currently on this branch, `SCALED_HA=true` won't be triggered on the next release for pre-upgrade deploys, we would unintentionally set `config.HA=true` for UAA in testing after the next release. Instead, I've recommended that `CUSTOM_UAA_SIZING=true` be set independent of [whether the deploy detects it's a pre-upgrade deploy requiring the pxc workaround](https://github.com/SUSE/cf-ci/blob/ffe344c2f46c976ad949a0ba0d11bec7a78cde53/qa-pipelines/tasks/cf-deploy.sh#L18) (as that will be unnecessary after the 1.5 release

qa-pipelines/tasks/cf-deploy.sh
```diff
@@ -14,12 +14,17 @@ pxc_pre_upgrade() {
     fi
 }
 
+# For now we will keep on using custom sizing for UAA.
+# Until CATs failures issue is addressed.
+export CUSTOM_UAA_SIZING=true
```
HartS: :+1:

qa-pipelines/tasks/cf-upgrade.sh
```diff
@@ -52,14 +52,19 @@ pxc_post_upgrade() {
...
 
+# For now we will keep on using custom sizing for UAA.
+# Until CATs failures issue is addressed.
+export CUSTOM_UAA_SIZING=true
+
+# We can remove the custom scf sizing after 1.5 release.
 if pxc_post_upgrade; then
-  # Need custom sizing for since config.HA_strict=false is not available for UAA charts.
-  export SCALED_HA=true
+   export CUSTOM_SCF_SIZING=true
```
HartS: 3-space punctuation

qa-pipelines/tasks/cf-upgrade.sh
```diff
@@ -80,25 +85,15 @@ if [[ "${EMBEDDED_UAA:-false}" != "true" ]]; then
...
+set_helm_params # Resets HELM_PARAMS.
+set_scf_params # Adds scf specific params to HELM_PARAMS.
 
+# Explicitly setting mysql count to 1 for pxc upgrade testing for scf.
 if pxc_post_upgrade; then
-  #HELM_PARAMS+=(--set=config.HA_strict=false)
   HELM_PARAMS+=(--set=sizing.mysql.count=1)
 fi
 
-if [[ "${EMBEDDED_UAA:-false}" != "true" ]]; then
```
HartS: Move these back for right now because we don't have a good way of checking the task type from here right now (need refactoring)

qa-pipelines/tasks/cf-deploy.sh
```diff
@@ -118,6 +113,7 @@ if pxc_pre_upgrade; then
     helm upgrade --reuse-values uaa ${CAP_DIRECTORY}/helm/uaa/ \
```
HartS: Placing `--reuse-values` on a new line would be a bit nicer

qa-pipelines/tasks/cf-deploy.sh
```diff
@@ -118,6 +113,7 @@ if pxc_pre_upgrade; then
...
         --namespace "${UAA_NAMESPACE}" \
         --timeout 600 \
+        --set "secrets.INTERNAL_CA_CERT=$(get_uaa_ca_cert)"
```
HartS: We don't need this in the uaa upgrade

qa-pipelines/tasks/cf-deploy.sh
```diff
@@ -127,6 +123,7 @@ if pxc_pre_upgrade; then
     helm upgrade --reuse-values scf ${CAP_DIRECTORY}/helm/cf/ \
         --namespace "${CF_NAMESPACE}" \
         --timeout 600 \
+        --set "secrets.UAA_CA_CERT=$(get_uaa_ca_cert)"
```
HartS: backslash

HartS: Prefer removing this; reuse-values makes it ok to not include it, and when we go back to testing embedded UAA we don't want the get_uaa_ca_cert function called because it will break

qa-pipelines/tasks/cf-deploy.sh
```diff
@@ -5,8 +5,26 @@ set -o nounset
...
 source "ci/qa-pipelines/tasks/lib/klog-collection.sh"
 
-set_helm_params # Sets HELM_PARAMS
-set_uaa_sizing_params # Adds uaa sizing params to HELM_PARAMS
+pxc_pre_upgrade() {
+    if [[ -n "${CAP_BUNDLE_URL:-}" ]] && [[ "${HA}" == true ]]; then
+        if semver_is_gte 2.17.1 "$(helm_chart_version)"; then
+            return 0    
+        fi
+        return 1
```
prabalsharma: print a statement here before returning 1

PR #276, opened 2019-08-01
Open cats yamls for write and not append, allows multiple runs

viccuad:
This fixes creating the yamls for `acceptance-tests-brain` test option,
so one can run the task several times and get the same files. If not,
yaml contents concatenate and give you an incorrect yaml.

It's just https://github.com/SUSE/cf-ci/pull/271 but against develop.

### DIFF
--- a/qa-pipelines/tasks/run-test.sh
+++ b/qa-pipelines/tasks/run-test.sh
@@ -133,7 +133,7 @@ if [[ ${TEST_NAME} == "acceptance-tests-brain" ]]; then
             if doc["kind"] == "Pod"
                 File.open("${test_pod_yml}", 'w') { |file| file.write(doc.to_yaml) }
             else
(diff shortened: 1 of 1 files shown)


HartS (changes requested):
The non-pods yaml file which is constructed contains multiple documents for other kubernetes resources such as ServiceAccount, RoleBinding, Role, etc. so overwriting the file each time another document is encountered will fail to create all necessary resources to run this test suite. This task should always run in a new container, so there shouldn't be an issue with appending to this file, but if running the script manually, you can delete the file after every run. If you *do* need a change in the script itself that allows multiple re-runs, you can add a line at the end to delete the non-pods yaml file

viccuad:
Closing this, then.

In my opinion, it would nice to have an idempotent approach that allows to apply the scripts several times without problem.

PR #278, opened 2019-08-02
Add script prepare-caasp4.sh to prepare CaasP deployments for CAP

viccuad:
prepare-caasp4.sh takes an existing CaasP4 deployment via kubectl, and creates the correct configmap, installs helm, an external storageclass with nfs, and the qa-sa config.

It also allows qa-tools/create-qa-config.sh to be run outside of a master node.

### DIFF
--- a/qa-tools/prepare-caasp4.sh
+++ b/qa-tools/prepare-caasp4.sh
@@ -1,18 +1,27 @@
 #!/usr/bin/env bash
 set -euo pipefail
 
-usage() {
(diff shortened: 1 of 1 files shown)


greygoo:
LGTM

viccuad:
Fixed bugs found by @greygoo.

viccuad:
Refactored, now the script receives env vars and fails if they are not set.
Ready for merging, I can squash the commits.

PR #283, opened 2019-08-04
Load test script for cf app spring-music

satadruroy:
Can be reused for soak as well.


HartS:
Do we want this app to have a manifest.yml?

satadruroy:
> Do we want this app to have a manifest.yml?

The cf sample app spring-music already has a manifest. 

https://github.com/scf-samples/spring-music

PR #287, opened 2019-08-14
remove k8s version for aks tf, fixes bugs

prabalsharma:
remove k8 version for AKS terraform
add enable_cf_usb_test to is_upgrade_pipeline variable
cf-get-commit-id needs to be out of conditions

### DIFF
--- a/cap-terraform/aks/variables.tf
+++ b/cap-terraform/aks/variables.tf
@@ -9,6 +9,7 @@ variable "location" {
 variable "node_count" {
     default = "1"
 }
+
 variable "machine_type" {
     default = "Standard_DS3_v2"
 }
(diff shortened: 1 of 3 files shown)


cap-terraform/aks/variables.tf
```diff
@@ -21,9 +21,6 @@ variable "cluster_labels" {
     type = "map"
 }
 
-variable "k8s_version" {
-    default = "1.13.7"
-}
 variable "disk_size_gb" {
     default = 60
 }
```
bikramnehra: Not exactly part of this change but can you add missing newline after:

```
variable "node_count" {
    default = "1"
}
```

PR #309, opened 2019-09-10
Disable autoscaler in default config

HartS:
This is in response to a race condition which requires the user to
manually run a workaround script when encountered. See
https://github.com/SUSE/scf/pull/2680 for more information


prabalsharma:
Closing this. We will handle this with a nightly CI with disabled autoscaler and credhub.

PR #312, opened 2019-09-16
Adding SCF v3 build pipelines

bikramnehra:
The PR intends to add code for v3 build pipelines. Also, it moves source control for SLE https://github.com/cloudfoundry-incubator/cf-operator-ci/tree/master/pipelines/release-images-cf-deployment pipeline over to cf-ci.

### DIFF
--- a/build-pipelines/generate-pipeline.rb
+++ b/build-pipelines/generate-pipeline.rb
@@ -34,4 +34,4 @@ class PipelineDeployer
     end
 end
 
-PipelineDeployer.new.render *ARGV.first(6)
+PipelineDeployer.new.render *ARGV.first(3)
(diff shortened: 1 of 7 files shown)


build-pipelines/generate-pipeline.rb
```diff
@@ -0,0 +1,37 @@
...
+        # Render the template
+        filename = File.join(__dir__, template_file)
+        template = ERB.new(File.read(filename))
+        template.filename = filename
+        result = YAML.load template.result(eval_context(config_file, secrets_file))
+        puts result.to_yaml
+    end
+end
+
+PipelineDeployer.new.render *ARGV.first(6)
```
HartS: This should be `first(3)` right?

PR #318, opened 2019-09-26
fix terraform pools

HartS:
- Ensure terraform pipelines use the correct pool branch/path and ignore
  pool setting (if provided)
- Parameterize tf pipeline locations
- Briefly describe how to deploy terraform pipelines in README


prabalsharma:
User should be able to use any pool they want. So if I choose prabal-eks pool for terraform-aks, I should be able to do it.
Also, removing config-terraform-aks/gke.yaml is kind of an exception from other config files. I think ./set-pipeline usage should be same for all types of deployments be it terraform or non-terraform.

prabalsharma:
if the issue is; if someone don't specify pools it uses pb pool. The solution should be that CI throws an error if no pool is specified by a user.

PR #321, opened 2019-10-01
dns for embedded uaa when lb is used

prabalsharma:


qa-pipelines/tasks/lib/azure-aks.sh
```diff
@@ -46,22 +46,41 @@ azure_dns_clear() {
...
+        "hostname": .status.loadBalancer.ingress[0].hostname
+       }
+    ]'
+}
+uaa_lb_info() {
+    kubectl get svc -n $namespace -o json |
+    jq -c '
+    [
+      .items[] | select(.spec.type == "LoadBalancer") |
+      select(.metadata.name == "uaa-uaa-public") | {
```
mook-as: Shouldn't we just always pass in the LB name, instead of having this weird mostly-duplicated function?
Or at least, call `lbs_info` and filter the results?

qa-pipelines/tasks/lib/azure-aks.sh
```diff
@@ -73,9 +92,14 @@ azure_check_lbs_ready_in_namespace() {
...
     local namespace=$1
+    local lb_name=${2:-""}
     local count=0
-    while ! azure_check_lbs_ready_in_namespace $namespace; do
-        sleep 30
+    while ! azure_check_lbs_ready_in_namespace $namespace $lb_name; do
+        if [[ -z "$lb_name" ]]; then
+            sleep 30
+        else
+            sleep 180
```
mook-as: Why is the timeout different here? Needs a comment at least. (Basing it on "if the lb has a name" is weird too… why isn't it just a separate parameter with a default value?)

qa-pipelines/tasks/lib/azure-aks.sh
```diff
@@ -46,22 +46,41 @@ azure_dns_clear() {
             --name ${matching_record_set}
     done
 }
+lbs_info() {
+    kubectl get svc -n $namespace -o json |
```
HartS: namespace should be set in the function, this is only going to work if it's called from a place where the `namespace` variable is already set

qa-pipelines/tasks/lib/azure-aks.sh
```diff
@@ -46,22 +46,41 @@ azure_dns_clear() {
...
+      .items[] | select(.spec.type == "LoadBalancer") | {
+        "svc": .metadata.name,
+        "ip": .status.loadBalancer.ingress[0].ip,
+        "hostname": .status.loadBalancer.ingress[0].hostname
+       }
+    ]'
+}
+uaa_lb_info() {
+    kubectl get svc -n $namespace -o json |
+    jq -c '
```
HartS: we should probably drop the `-c` here and above (I know this was already in the script)

PR #337, opened 2019-11-18
Fixing wrong pipeline trigger issue on `kubecf`

bikramnehra:
This fixes the issue where `master` was not being updated correctly before opening PR to update buildpacks.


build-pipelines/buildpacks/pipeline.yml.erb
```diff
@@ -101,8 +101,6 @@ resources:
   source:
     uri: <%= kubecf_repo %>
     branch: <%= kubecf_branch %>
-    paths:
```
f0rmiga: I think we still want this here.

PR #350, opened 2019-12-16
Switching stemcell to SLE15

bikramnehra:
This PR intends to switch stemcell version to SLE15.

Addresses: https://github.com/SUSE/kubecf/issues/255


kubecf-build-pipelines/cf-deployment/config.yml
```diff
@@ -3,17 +3,17 @@ cf-deployment-tags:
   - v9.2.0
   - v9.5.0
 ci-repo: https://github.com/suse/cf-ci
-ci-branch: develop
+ci-branch: bisingh/sle15-stemcell
```
f0rmiga: This is wrong.

f0rmiga (changes requested):

f0rmiga (approved):

PR #351, opened 2019-12-19
Clean up list of pipelines

thardeck:
Where does https://concourse.suse.de/teams/main/pipelines/fissile-develop come from? It does not seem to be the develop pipeline from bosh-linux-stemcell-builder-ci.

We do not have develop images anymore, but this seems to be more a fissile-ci pipeline.

### DIFF
--- a/deploy-all
+++ b/deploy-all
@@ -85,28 +85,6 @@ ensure
   run 'make', 'post'
 end
 
-Dir.chdir('../cf-buildpacks-ci') do
-  def update(*args)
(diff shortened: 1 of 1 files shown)


mook-as (approved):
`fissile-develop` is the `develop` branch of `fissile` (as opposed to the `master` branch, which we were using in SCF before we… kind of got lazy and stopped doing that for fissile)

mudler (approved):

PR #353, opened 2020-01-15
Remove timeout for upgrade app downtime monitoring

HartS:
If the app isn't back online, we have problems with the upgrade, so we
should keep monitoring until the cf-upgrade task timeout hits (or helm
upgrade timeout if applicable). Monitoring should not complete
successfully unless the app is back up.

### DIFF
--- a/qa-pipelines/tasks/cf-upgrade.sh
+++ b/qa-pipelines/tasks/cf-upgrade.sh
@@ -22,10 +22,6 @@ monitor_url() {
       last_state=$new_state
     fi
     ((++count))
-    if [[ ${count} -gt 10800 ]]; then
-      echo "Ending monitor of ${app_url} due to timeout"
-      break
-    fi
     sleep 1
   done
 }


prabalsharma:
so the upgrade task timeout is 5hours. I think instead of removing timeout from monitoring we should exit 1. This will give us more info more quickly

`    if [[ ${count} -gt 10800 ]]; then	
      echo "Ending monitor of ${app_url} due to timeout"	
      exit 1
    fi`

prabalsharma (changes requested):
so the upgrade task timeout is 5hours. I think instead of removing timeout from monitoring we should exit 1. This will give us more info more quickly

```   if [[ ${count} -gt 10800 ]]; then	
      echo "Ending monitor of ${app_url} due to timeout"	
      exit 1
    fi

HartS:
@prabalsharma it's true that downtime really shouldn't be that long, but in the event of slow image pulls combined with other defects which might cause excessively long upgrades (and therefore, app downtime), it may be valuable to know where things are going wrong, rather than timing out the app monitoring.

Also, since this is backgrounded, exiting with status 1 won't actually trip errexit and fail the test, so the timeout is irrelevant; I don't see any reason not to continue monitoring as long as the application is not ready post-upgrade

prabalsharma (approved):

PR #365, opened 2020-02-26
Remove vancouver pipelines from deploy-all

thardeck:
The instance is down since month, so I guess it is safe to remove it.


mook-as (approved):
LGTM, getting @HartS to take a look too because I haven't touched the pipelines in a while.

thardeck:
@HartS any update on this topic?

HartS (approved):

PR #368, opened 2020-03-19
fix: chop off pxc from the cf-deployment releases

f0rmiga:
Removing pxc BOSH release as it fails to build: https://concourse.suse.de/teams/main/pipelines/kubecf-cf-deployment/jobs/build-v12.36.0/builds/5.


f0rmiga:
This is part of https://github.com/cloudfoundry-incubator/kubecf/issues/473.

bikramnehra (approved):
LGTM. Assuming its testing already.

PR #370, opened 2020-03-24
Add missing version field

bikramnehra:
This is breaking kubecf build as the `version` field is used in order to build the final URL to be pulled.
See: https://github.com/cloudfoundry-incubator/kubecf/pull/546


HartS (approved):
LGTM

PR #376, opened 2020-07-31
Bump CATS to 0.0.16

mook-as:
Actually switch the cf-acceptance-tests to use the correct branch for cf-deployment 13.9.0.


andreas-kupries (approved):
:+1:

